PR #13, opened 2017-11-06
Port from Source1 addressing the following issue: Fixed account chara…

drk84:
…cter creation not creating beards.

See:
https://github.com/Sphereserver/Source/commit/78c9a30b7c1a7fbd7726db8a247369148e8619f1#diff-824b80dc08bda51376a5a5aad994302f


cbnolok:
Closing because the bug was related to the character creation packet itself, which will be fixed in a new commit (which will also include part of your pull request).

PR #87, opened 2018-05-10
Fixed: Invis (statf_invisible) chars being revealed when others chars…

azmanomer:
… walk over him.


azmanomer:
b2f41361ea106527721c78e27adcc273c34072c8 reverted invis change from this commit

azmanomer:
Btw i just check OSI behaviour and it seems when u are under invis spell and someone step on you. You are not revealed and the person step on you didnt receive any message.

PR #107, opened 2018-05-25
The following message: DEFMSG_COMBAT_ATTACKO (*%s is attacking %s!*) …

drk84:
will use the EMOTE_DEF_COLOR setting found in sphere_msgs.scp instead of always use the gray color hue.

Moved methods Skill_Snoop_Check, Skill_Snooping, Skill_Stealing from CCharFight.cpp to CCharSkills.cpp

Moved the code related to the Focus skill from method Stat_Regen (CCharStat.cpp) to new method Skill_Focus (CCharSkills.cpp)


cbnolok:
Skill_Focus needs to return an int (like the other Skill_* methods), because it needs to be able to return signed values:
-SKTRIG_QTY = no chance. and not a crime
-SKTRIG_FAIL = no chance and caught.
0-100 = difficulty = percent chance of failure.

drk84:
If i return -SKTRIG_QTY i return -12 right? If yes, i need to find a different way to implement the function :o

cbnolok:
It's just a conventional exit value for the skill function, like
-SKTRIG_ABORT, (they're all negative afair) and it tells that the
method/skill stage ended, in which state and why. You can just test if
Skill_Focus(..) > 0, then sum the return value

Il 26 mag 2018 12:41, "drk84" <notifications@github.com> ha scritto:

If i return -SKTRIG_QTY i return -12 right? If yes, i need to find a
different way to implement the function :o

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/Sphereserver/Source-experimental/pull/107#issuecomment-392253013>,
or mute the thread

PR #110, opened 2018-05-28
 Possible re-implementation of old AR shield formula.

drk84:
This works only with Combat Elemental Engine disabled
Instead of gaining  a flat 7% of the Shield Armor value the Parrying skill of the character is taken in consideration according to the formula on
http://web.archive.org/web/20000306210936/http://uo.stratics.com:80/parr.htm

Take note  that by default the ARMOR value of the Shields item is commented in the scripts, so you need to uncomment it.

### DIFF
--- a/Changelog-56d-Nightlies.txt
+++ b/Changelog-56d-Nightlies.txt
@@ -984,3 +984,16 @@ Redeed now creates the deed before calling the @Redeed trigger and the deed gets
 	the @Fail, @SkillFail, @Start, @SkillStart triggers to be fired at each hit.
 	- Fixed: NPCs doing extra AI actions (wandering, food, etc checks) after each hit in combat. This caused sometimes the current combat action to be resetted
 	and then forced again, erasing the NPC combat data, which ultimately results in the NPC hitting twice in a few tenths of seconds.
+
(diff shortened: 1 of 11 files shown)


drk84:
A new value in CombatParryEra like:
// 080  Enable  Shields AR scaling based on parrying 

I think it's T2A, the page is from 2000 and LBR was released on 2002, while AOS on 2003.
Hovewer by looking through the website, the parrying it's more complicated :S 
While the displayed AR on the status bar is ok, the AR given from the shield is applied only when you block the blow!
And if it's melee blow you block 
Damage reduced for melee attacks = (Base AR) /2
While if it's a ranged attack you block full damage (like now)
Damage reduced for missile attacks = Base AR

_"If you do not parry the blow, the damage goes through as normal. But if you are successful, your shield will absorb some of the damage thus reducing the attacker’s blow. The amount of damage subtracted depends on the type of shield you are using. A shield will absorb an amount of damage equal to half the base AR of the shield for melee damage, and the full base AR for missile damage. "_

drk84:
So it seems like that displayed AR is just for "show", while what it actually matters for blocking the damage is the base AR of the shield.

PR #115, opened 2018-07-23
LocationEffect, plus the old parrying that i deleted for error and a meditation/spell tweak.

drk84:
Added: LocationEffect command, similar to the EFFECT command but instead of an object it takes a terrain location as a target.(Thanks to XuN for the help)

 Some of the new spells use effects that are placed  on the ground (not field items) for simulating an area of effect.
 Usage: locationEffect x,y,z,type,itemid,speed,loop,explode,color,render
 Examples:
 //Simulate the Consacrate Weapon Chivalry spell effect, a sword appearing from the ground and going upward.
 [FUNCTION testLocationZ]
 locationEffect <p.x>,<p.y>,<height>+1,0,i_sword_long,1,0,0,colors_red,3

 //Simulate a field of energy centered at the target.
 //argv0 x offset, argv1 y offset
 [FUNCTION testLocationField]
 if <isempty <args>>
  args = 1,1
 endif
 local.pX = <p.x>
 local.pY = <p.y>
 local.pZ = <p.z>
 FOR X <eval <local.pX> - <argv0>> <eval <local.pX> + <argv1>>
  FOR Y <eval <local.pY> - <argv0>> <eval <local.pY> + <argv1>>
   locationEffect <local.x>,<local.Y>,<local.pZ>+<argv2>,2,i_fx_energy_field,1,0,0,colors_red,3
  ENDFOR
 ENDFOR
  Notes:
  If the effect Type is 1 (Lightning) or 3 (Effect at object level) the effect will not be played, only effect 0(bolt) and 2(ground level) seems valid, probably
  cause they don't accept a map point.

- Added: Spells with the LAYER property set to >= LAYER_SPELL_STATS will automatically equip the spell memory item found in the RUNE_ITEM property.
 Currently, if you want to make a custom spell (or scripts the chivalry, mysticm, spelleweaving spells) you had to manually equip the memory item
 causing the triggers @EffectAdd  and @EffectRemove to never be fired.
- Changed: Meditation skill now take in consideration the EFFECT property of the skill, if the EFFECT property is not present the default value of 1 is used.
- Added: Added new flag value 080 for CombatParryingEra, setting this flag will allow to use the following parrying skill behaviour:
 The flag will enable the old AR parrying formula, instead of gaining  a flat 7% of the Shield Armor value the Parrying skill of the character is taken in consideration according to the
 following formula:
 Displayed AR = ((Parrying Skill * Base AR of Shield) ÷ 200) + 1
 The maximum AR reachable is: Base AR of Shield / 2.
 More information: http://web.archive.org/web/20000306210936/http://uo.stratics.com:80/parr.htm
 Notes
 By default the ARMOR value of the Shields item is commented, so you need to uncomment them.
 The formula works only with Combat Elemental Engine disabled.
 Setting or unsetting this flag while the server is running will update the character armour value only when the Parrying skill changes or the shield item is equipped/unequiped.


coruja747:
github notified me about this pull request, it seems to be a nice change, but since I'm not developing this experimental branch maybe it's better wait for someone else approve :P

anyway, would you mind if I port it to main branch?

drk84:
Yes you can of course :D

2018-07-23 21:20 GMT+02:00 Coruja <notifications@github.com>:

coruja747:
thx, I rewrote the code in a bit different way, using a shorter name (EFFECTP instead EFFECTLOCATION) to make it standard with others P-based functions (ACTP, TARGP, etc), and it also uses full src/dest P instead only dest P (this will make the function able to handle more effects, like meteors falling from the sky, fire going from nowhere to the char, etc)

feel free to take something if you want:
https://github.com/Sphereserver/Source/commit/882b049f870cc7a18bcdc9883937c520afb936c3
https://github.com/Sphereserver/Source/commit/94c2542cd37d3591779304b485db22d095809d22

Issue #221, opened 2019-02-10
@MurderMark ---> return 1 (feature request)


AtillaG1:
to be able to stop any default behaviour for our own kill count system and our rules to gain kill.
is it possible please?
Maybe can be done with @MurderDecay ---> return 1 too

AtillaG1:
Thanks to Julian, nice solution would be:
In CCharNotoriety.cpp
start line 553.
`        // I'm a murderer !
        if ( !IsPriv(PRIV_GM) )
        {
            CScriptTriggerArgs args;
            args.m_iN1 = m_pPlayer->m_wMurders+1;
            args.m_iN2 = true;
            args.m_iN3 = false;

            if ( IsTrigUsed(TRIGGER_MURDERMARK) )
            {
                OnTrigger(CTRIG_MurderMark, this, &args);
                if ( args.m_iN1 < 0 )
                    args.m_iN1 = 0;
            }

            if ( args.m_iN3 != true )
            {
                m_pPlayer->m_wMurders = (word)(args.m_iN1);
                if ( args.m_iN2 )
                    Noto_Criminal();

                Noto_Murder();
            }

            NotoSave_Update();
        }`
argn3=1 in @murdermark ---> ignore any default behaviour about kill gain and criminal

Issue #219, opened 2019-02-06
When closing a dialog from a memory item under @timer the dialog doesnt close all the way


soulless320:
if you close the dialog for a player under @timer it will still show the dialog as open when doing .show isdialogopen

![image](https://user-images.githubusercontent.com/17236327/52316920-b9729d00-298b-11e9-862d-32b375f9dd7c.png)

GitMadKing:
use cont.closedialog (if equipped)
not only closedialog

soulless320:
It's already closing it correctly, the issue however is that it still shows on 'isdialogopen' that it's not closed out

`on=@timer
if (<more1> <12) && (<def0.welcome_dialog_animated>)
  more1 += 1
  trysrc <link> dialogclose d_welcome
  trysrc <link> dialog d_welcome
endif
timerms=75
return 1`

soulless320:
Fixed, thank you Shiryux!

Issue #57, opened 2018-05-06
[PORT] 3066341740896269766206db690f76a3785290b1, Fixed: Attack/Kill command on pets allowing select the pet itself as target, making it attack his owner


nefthon:
Revision: https://github.com/Sphereserver/Source/commit/3066341740896269766206db690f76a3785290b1
Author: Coruja coruja@outlook.com
Date: 30/06/2016 00:41:34
Message:
Fixed: Attack/Kill command on pets allowing select the pet itself as target, making it attack his owner

PR #225, opened 2019-02-11
quick fix for issue #221, #57 and #219

shiryux:
Issue #221 
Added ARGN3 to @MurderMark which if is set to anything but 0, it will avoid adding the kill and marking the char as criminal/murder. Note that I left out NotoSave_Update(); in case some scripter forgets to use notoupdate after doing their custom checks.


shiryux:
no idea why but it merged both commits,
it is supposed to solve the issue #57 which is a port from 56d
and issue #219, Dialog_Close first argument is "this" and it should be the char from pClientSrc, in case you call src with trysrc,

cbnolok:
Can you add some info to the changelog?

shiryux:
done! (no idea why is pulling all my commits, first time I do this)

PR #266, opened 2019-04-10
Fixes for ContextMenu and dynamic gump lines.

JulianUO:
Okay, so this is the revision (by the way, you forgot to switch years haha)

```
- Fixed: Feature sending to unencrypted clients (mostly with Razor) using ReportedCliver as reference instead of CliVer.
- Fixed: Text index for dynamic dialog lines, such as dhtmlgump, they were starting at random ranges instead of starting at 0.
- Modified: Hardcoded the train and tame function for Context Menu (or Popup Menu), this way you can remove e_Trainer from every NPC.
```

First entry: When connecting with razor seems that tmVer is 0 at PacketCharacterList, so FeatureFlags are gonna be according to client 0. Instead, I used the reported cliver which is tmVerReported. I can see the client version on that variable and the server will send the correct feature flags.

Second entry: I don't know if that is the best solution, the index was starting at 1400000 or some weird number when using uint. In the default client this works fine, but when using ClassicUO, the client crashed since it was expecting and index from 0~amount of text. Anyway, switching to int and using %d as argument fixed it.

Third entry: Some commit I took from 56, can't find when, but the default script pack isn't using e_trainer anymore and if someone wants to use that script pack and this sphere version, they won't have any train button on the Context Menu.

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -1582,7 +1582,7 @@ WARNING: the following change could break some hackish expedient used in scripts
 24-03-2019, Julian
 - Modified: @HitParry to include local.Damage, which contains the raw amount of damage before any parrying modification.
 
-31-03-2018, Nolok
+31-03-2019, Nolok
 - Fixed: Swimming NPCs could pass under items even if their z wasn't high enough.
 - Fixed: NPCs could always walk inside places with a ceiling too low for them to fit below.
 - Fixed: The absence of MT_WALK didn't stop NPCs from walking on land.
@@ -1590,7 +1590,7 @@ WARNING: the following change could break some hackish expedient used in scripts
 - Fixed: BTFish encryption not working (Issue #256).
 - Fixed: During character creation, invalid professions sent by the client will now be ignored.
 
-05-04-2018, Nolok
+05-04-2019, Nolok
 - Added: trigger @CreateLoot. It fires when the NPC dies, if it is allowed to have a loot (not a player, not conjured, no DEATHF_NOLOOT).
 	The main advantage of using this trigger in place of @NPCRestock and @Create is that items will be created only when they are required,
 	so that a big number "useless" items won't be stored in the save files (= smaller save files), also GMs wouldn't be able to see what a monster is going to drop before it dies.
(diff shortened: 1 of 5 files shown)


cbnolok:
Can you fix please the compilation errors on Linux? See https://travis-ci.org/Sphereserver/Source-experimental/builds/518498501?utm_source=github_status&utm_medium=notification
Also the warnings:
```
/home/travis/build/Sphereserver/Source-experimental/src/game/clients/CClientEvent.cpp: In member function ‘void CClient::Event_VendorSell(CChar*, const VendorItem*, uint)’:
/home/travis/build/Sphereserver/Source-experimental/src/game/clients/CClientEvent.cpp:1343:36: warning: for increment expression has no effect [-Wunused-value]
  for (uint i = 0; i < uiItemCount; i)
```
Have you tried if using %u instead of %d (since CDialogDef::GumpAddText returns an unsigned integer) keeps that working?

JulianUO:
The warning on Event_VendorSell comes from the MORE2/NEWGOLD commit.
Weird that no one reported bug with vendors?

I think the linux warnings are fixed, plus fixed the casting, it was using static_cast instead of ( ). I have no way to test on linux right now. Also, switched to uint, just tried and it works fine, my bad.

cbnolok:
Yeah you're right, that warning was my fault :D
Linux compilation is fixed now, before i merge can i ask if you can test if changing `%" PRIuSIZE_T` with %u (instead of %d) still works? Since that value is a uint, when it reaches a value > INT_MAX, %d displays that value as a negative number

cbnolok:
Yeah you're right, that warning was my fault :D
Linux compilation is fixed now, before i merge can i ask if you can test if changing `%" PRIuSIZE_T` with %u (instead of %d) still works? Since that value is a uint, when it reaches a value > INT_MAX, %d displays that value as a negative number

cbnolok:
Yeah you're right, that warning was my fault :D
Linux compilation is fixed now, before i merge can i ask if you can test if changing ```%" PRIuSIZE_T``` with %u (instead of %d) still works? Since that value is a uint, when it reaches a value > INT_MAX, %d displays that value as a negative number

cbnolok:
Yeah you're right, that warning was my fault :D
Linux compilation is fixed now, before i merge can i ask if you can test if changing ```%" PRIuSIZE_T``` with %u (instead of %d) still works? Since that value is a uint, when it reaches a value > INT_MAX, %d displays that value as a negative number

JulianUO:
Done! Sorry for the ammount of commits, seems ready for merge now.

cbnolok:
That's perfect, thanks!

PR #313, opened 2019-08-05
Fixed (i think): Being stuck on a web indefinitely, unless the web wa…

drk84:
…s created by a spider. Default stuck value is now around 6 seconds.

Now, after the STUCK memory item expires if the character continues to move towards the web it will get stuck again, while if move in another direction he will be free from the web.

 The web item didn't have a timer set, that means the character  would get a  STUCK memory item that never expired because the STUCK memory item get his timer value from  the web  the character stepped on.

But i remember that the old web behaviour was different: you got stuck on the web until you slowly "break it" by trying move, it seems now that being frozen blocks every attempt to move?

- Fixed: Purple potions MoreX (countdown to explosion) always ignored when set, default countdown value is 4 seconds.

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -1847,7 +1847,7 @@ WARNING: Important change!
 30-07-2019, Nolok
 - Fixed: Tooltip data not being sent right after an item was added to a container, in that case a random tooltip was shown for the item.
 
-02-07-2019, Nolok
+02-08-2019, Nolok
 - Fixed: Connections to the HTTP server were always blocked, even if legit.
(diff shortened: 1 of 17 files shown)


cbnolok:
Could it be that in the past ststf_freeze was not set, and now it is? In that case the client would block the player actions

drk84:
By removing the statf_freeze it works similar with the "fix" i sent,  continuing to move towards the web will weaken the web but moving in another direction will allow you to move freely

cbnolok:
Mmm then wait a sec, yesterday i fixed part of the problem but your change is still necessary, maybe with my fix you could even remove the statf_freeze flag and restore the old behavior.

PR #315, opened 2019-08-13
Possible Fix?  Items not buyable from Vendors.

drk84:
It seems that struct items was not initialized? I don't know if it's the correct way, i just checked this:

http://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Initializing-Structure-Members

### DIFF
--- a/src/network/receive.cpp
+++ b/src/network/receive.cpp
@@ -726,7 +726,7 @@ bool PacketVendorBuyReq::onReceive(NetState* net)
 		return true;
 	}
 
-    VendorItem items[MAX_ITEMS_CONT] = {};
+    VendorItem items[MAX_ITEMS_CONT] = {0};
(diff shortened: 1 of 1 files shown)


cbnolok:
The previous syntax should be okay afaik, it's a cpp initializer list, when it's empty the data will be zero initialized. Maybe the problem lies in another change?

drk84:
Pratically   i have noticed  that without the {0},  all the elements of VendorItem items  have the m_serial property always set to the maximum value of dword (4294967295)
immagine

and thus skipping this check ( i don't know why the serial is compared to 0)
immagine

cbnolok:
Oh yes, because the empty brackets default initialize the elements (thus, using the constructor), while with 0, it's a zero initialization of the memory. I used empty brackets because i supposed that isvaliduid was used, instead of checking its val against zero, but it's fine anyways

Issue #202, opened 2019-01-15
You can commit custom house changes even they are not proper


azmanomer:
There is a bug when you try to commit not proper changes for custome house like below;
Steps to reproduce this;
Do it like in the picture below and tryo to commit
![resim](https://user-images.githubusercontent.com/13722155/51185776-12996480-18e9-11e9-9a3b-35110b76d344.png)
it ll say "Illegal floor! House not legal, not commiting!"
Then remove only 1 tile from those orange tiles and when you do that the whole floor will dissappear then press commit again. It ll be like this ;
![resim](https://user-images.githubusercontent.com/13722155/51185961-89366200-18e9-11e9-8fbe-104409118f4c.png)

azmanomer:
https://vimeo.com/312764388

AtillaG1:
cant commit https://i.gyazo.com/f720c9d63113158450b3128cd157fe98.png

not able to reproduce it, you sure it isnt script problem ?

azmanomer:
yep its only happening if its fresh floor :)

AtillaG1:
[![Screenshot from Gyazo](https://gyazo.com/9a997d865123df71c28d90d3ed4b4ee8/raw)](https://gyazo.com/9a997d865123df71c28d90d3ed4b4ee8)x

cbnolok:
It should work now

Issue #323, opened 2019-09-06
custom housing


AtillaG1:
also on vanilla scps
[![Screenshot from Gyazo](https://gyazo.com/70b201fff3675f894b4740c2a1693939/raw)](https://gyazo.com/70b201fff3675f894b4740c2a1693939)

cbnolok:
Shouldn't that be the default behavior? Isn't that happening also on 56c or d?

AtillaG1:
on corujas build it is doing everything OK. Even the stairs are different as we are used to.

soulless320:
Yeah, deleting one tile shouldn't remove the whole floor, it not only removes the entire floor but it removes everything on the level as well, this includes walls and doors. :O

cbnolok:
Clearly it shouldn't remove walls and the doors, though i'm not sure that it shouldn't remove the whole floor. I guess that on 56d issue #202 is still present (if the whole floor isn't removed there), and in the attached video you can see that (at least in that case) the client removes the whole floor (that bug was in fact caused by the client removing the whole floor, but sphere internally removing only the targeted tile).

cbnolok:
Actually, by testing more, it looks like the whole floor is removed only if it's invalid, gotta think about how to solve this now

AtillaG1:
missing floor is good.

PR #330, opened 2019-10-15
Updated custom multi AddStairs/RemoveItem

JulianUO:
Modified the loop to search through iStairID in AddStairs (though using a short, no idea if that's the correct naming convention), now it is similar to what you can see in RemoveStairs. The last argument seems to be related to AddMulti but couldn't find a reason to keep it.

RemoveItem was modified to not remove the entire floor no matter what, client doesn't seem to like that, the server might remove the floor but you can commit and see the whole invalid items placed on it or whatever was removed from view by the server. I removed that bunch of code, maybe it would be the best to put it between some #ifdef or leave it commented.

Anyway, these are the differences I found between the 56d and X versions.
Tested on latest client (7.0.69?). Everything seems to work fine.

### DIFF
--- a/src/game/items/CItemMultiCustom.cpp
+++ b/src/game/items/CItemMultiCustom.cpp
@@ -720,6 +720,58 @@ void CItemMultiCustom::RemoveItem(CClient * pClientSrc, ITEMID_TYPE id, short x,
     bool fComponentsChanged = false;
     auto& vectorComponents = m_designWorking.m_vectorComponents;
 	    
+	/*
+	if (fAllowRemoveWholeFloor && (uiPlaneAtZ > 1))
+	{
+		// If i'm removing a single floor tile and below there's a stair, this tile is invalid;
+		//  by removing this single invalid tile, the client actually removes the whole floor (even if it has different IDs), so we need to reflect that change here.
+		// It's not perfect though, because the client removes the whole floor if on the same row or column of the selected tile there's a stair below.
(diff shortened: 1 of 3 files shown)


cbnolok:
The floor removing thing was added to fix issue #202, further info about the necessity of that behavior is in issue #323 discussion

JulianUO:
I couldn't reproduce any of the bugs in #202 in this commit.

I just had this issue but couldn't reproduce it: 
1. Added walls of base floor. 
2. Filled first floor with stone pavers. 
3. Went to base floor and added stairs.
4. Went to first floor and couldn't see the stone pavers.
Everything restored after commiting.
I'll retest later to see if can reproduce it.

JulianUO:
I've tried and tried but could not reproduce any bug.
Not even the #202 bugs.

Before this pull request, when I removed an invalid tile in 2nd/3rd floor, the entire floor was removed from me, but when committed, the invalid tiles were there.

cbnolok:
Did you do those tests with an updated version? Because i'm running latest and by removing an invalid floor tile (actually sphere's selection is wider, because it considers invalid every floor tile which has a stair below, i couldn't do better without revolutionizing the code) in 3rd floor it removes the full current floor

JulianUO:
This is recording with the latest build (just downloaded from the automated builds)
https://streamable.com/m2pmn

cbnolok:
So do you confirm it's working with this pull request?

JulianUO:
Did some testing and everything worked fine with this pull request.
Anyway, I fear the same, I'll add another commit with the code commented before you can pull it.

Issue #241, opened 2019-02-27
New flag idea options for the ini


soulless320:
I was discussing with Drk and there appears to be some changes that were made in 56d that we think should be toggles in the ini, does anyone have any thoughts on these flags?

npcs not being interrupted during spellcasting when damaged, this should be a flag
  // Disturb magic spells (only players can be disturbed)
  maybe npc_spellcast_fizzle

npcs not getting bonus damage in  combat, this should be a flag
  // only players can have damage bonus
  perhaps  npc_combat_bonusdmg

and npcs seem to spam their spells one after the other till they are out of mana, there should perhaps be a setting for the average time delay between when npcs will cast their harmful spells
  maybe npc_spellcast_delay

soulless320:
just updating this. npc bonus damage complete. npc fizzle toggle in INI is being added.

The only thing left to address is NPCs spamming their spells.

soulless320:
Thanks for the additions!

PR #340, opened 2019-11-13
Possible solution for issue #241 and for the rock throwing spam and some other things

drk84:
- Added: COMBAT_NPC_BONUSDAMAGE Combat flag in the sphere.ini, enablng this flag  will allows the NPCs to gain the following modifier to combat damage:

   -Bonus from the CombatBonusStat and CombatBonusPercent properties:
     The default value for CombatBonusStat is STR.
     The default value for CombatBonusPercent varies according to the CombatDamageFormula 10% if CombatDamageFormula is 0, 20% if CombatDamageFormula is 1 and 30% if CombatDamageFormula is 3.
     See changelog of the 19-03-2015  from the 56c-Changelog.txt  revision file for more details on CombatBonusStat and CombatBonusPercent.
   -Modifier from the INCREASEDAM property (from -100 up to 100%)
   -Anatomy, Lumberjacking and Tactics if the CombatDamage formula is set to a value above 0.
   This should satisfy entry #241
- Added: BREATH.DAMTYPE, used for changing the type of damage of the breath. Default value is: DAM_FIRE.
- Added: THROWDAMTYPE, used for changing the type of damage of the thrown item. Default value is: DAM_PHYSICAL.
   Remember that NPCs need at least one the throwable item equipped or in the backpack.
- Added: Two new damage flags, DAM_BREATH and DAM_THROWN. These are added automatically to the BREATH.DAMTYPE and THROWDAMTYPE properties, i think they are useful for discriminating the source of damage in the @GetHit trigger.
- Changed: Veterinary now works with all tameable creatures: that means all NPCs with brain set to BRAIN_ANIMAL, BRAIN_DRAGON or BRAIN_MONSTER AND with a Taming skill above 0.
     Before Veterinary only worked with BRAIN_ANIMAL npcs. Remember that a NPC  is tameable when it has a Taming Skill above 0 and an Animal Lore skill set to 0.
- Changed: All NPCs with a Poisoning skill above 0 can now apply poison (if unarmed), before only creatures with BRAIN_MONSTER could apply a poison.
- Fixed: There was no timeout for the NPC action THROWING, resulting in a mass spam of throwing actions, the timeout is now 3 seconds.


cbnolok:
For veterinary, i'd add a check for humanoid bodies, in that case i'd make it use healing instead

drk84:
Yes it's better, also i can add the same check to Taming skill because you can tame an humanoid (human, elf or gargyole) if they have the Taming skill > 0 and AnimallLore == 0 :D

Issue #351, opened 2019-12-29
maxfollower and cast


alexrmrm:
When you have the maximum number of followers and try to castrate, use regents and mana, you should check before this as with the mana and regents.

Seokotov:
you can use check in script when casting if you wanna do it before...

alexrmrm:
it works well it can be closed

PR #361, opened 2020-01-24
Fix for Issue #351 and implemented feature for Issue #241

drk84:
Loss of reagents and mana when a summoning spell fail because Maxfollower property was reached. (Issue #351)
   You will still lose reagents and mana if ManaLossOnFail and ReagentLossOnFail are enabled in the ini.
   Note II:
   In this case i split the old Spell_Summon method in Spell_CanSummon and Spell_Summon:
   Spell_CanSummon creates the creatures and check if the summoned creature can be placed in the world, this happens before the consumption of mana and reagents.
   Spell_Summon  handles the placement of the creature in the world.

New ini flag DisplayArmorAsPercent. (Issue #241)
         When enabled, display, in the tooltip, the Armor of the item as a percentage of its full armor value, the percentage is based upon the body parts the item covers.
         Remember that the displayed value is just for "cosmetical" purpose and, by default, is not used when calculating combat damage, the full AR is the one used.
   In some cases, the Armor value displayed in the tooltip will not match with the sum of the AR of each piece of armor displayed in the status gump, this happens when an armor item cover location
   overlaps with another armor item cover location that as an higher value.
   Quick summary of layer body coverage:
   Only works when COMBAT_ELEMENTAL_ENGINE is disabled and FEATURES_AOS_UPDATE_B is enabled.
   PS:
   I made a static method named CChar::CalcPercentArmorDefence in CcharFight.cpp, then i call it in the AosTooltip_addDefaultItemData(CClientMsg_AOSTooltip) by using:
   CChar::CalcPercentArmorDefence(pItem->Item_GetDef()->GetEquipLayer());
   Maybe there is a better way?

Other minor fixes:
  - In Fight_Parry method (CCharFight.cpp) characters were not able to increase their parrying skill if the parrying or bushido value was under a certain threshold.(replaced <= 0 with < 0)
  - In GetSpellBook method (CCharStatus.cpp) characters were not able to cast the last spell in a spellbook. (replaced iSpell < max with, iSpell <= max)
  - Added: Missing layer LAYER_SPELL_Enchant and missing spell s_purge_magic.


cbnolok:
I think that Spell_CanSummon and Spell_Summon are misleading names, what do you say? It looks more suited to me something like Spell_Summon/Spell_Summon_Try and Spell_Summon_Place (also little typo: it's defense instead of defence). That said, it's very good work

drk84:
I renamed Sphere_CanSummon to Sphere_Summon_Try and Sphere_Summon to Sphere_Summon_Place (and fixed the typo in defence)

PR #364, opened 2020-02-05
Added missing features for Issue #241:

drk84:
New ini flag NPCCanFizzle, when enabled, a NPC can fizzle a spell when hit during combat.

TAG.NPCNoCastTill behaviour, this tag will not allow a NPC to cast a spell if tag.NPCNoCastTill > SERV.Time.

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -2121,3 +2121,12 @@ Must have event, even if it's empty, since it's applied by the source to generat
 25-01-2020, Nolok
 - Fixed: Invalid keywords in the [ADVANCE] block were likely to cause memory corruption.
 - Updated: SQlite to version 3.31.0.
+
+05-02-2020, Drk84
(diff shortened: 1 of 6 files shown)


cbnolok:
NPCCanFizzle is meant to work only for NPCs hitting other NPCs, right? Because it seems to not affect players hitting NPCs now. Also, i'd use a more descriptive name, because like that it seems that npcs don't fizzle in any case if this flag is off

drk84:
It should works when a player hits an NPC in combat and the NPC is trying to cast something, i tested it with a NPC casting a slow spell like flamestrike and it seemed to work if i remember well..

I can change the name to NPCCanFizzleHit or NPCCanFizzleWhenHit btw or something similar

cbnolok:
Then yes i'd change that to NPCCanFizzleHit or something similar but more specific if you prefer

drk84:
I renamed it on NPCanFizzleOnHit

Issue #389, opened 2020-03-28
Vendors don't sell anything


orishi:
Somewhere after this version:
SphereServer Version X1-Nightly [WIN64] by www.spherecommunity.net.
Compiled at Feb  2 2020 (15:25:10) [build 3222 / GIT hash c5cd3dece2bff7b011f08c40892556a7b23d79d2]

Vendors no longer sell anything. When you say "buy", they told you that they have no goods for sale.

Here is vendor script and templates:
https://pastebin.com/bLV2iELi

Looks like there is something wrong with @NPCRestock.
We tested this with Rastrero, he had the same problem. Reverting back to Feb 2 2020 version and everything works fine.

Client 7.0.20.0.

![idng](https://user-images.githubusercontent.com/2725562/77825492-fd81ad80-7109-11ea-8bf0-13c733681807.gif)

Tolokio:
looks like @npcrestock is not loading the ítems into the proper layer.

meaowR:
still happen in #3246

PR #392, opened 2020-03-31
Fixed restock issue #389.

xwerswoodx:


xwerswoodx:
Seems like because of using failed build, that pull request is failed too, so I will wait nolok to fix failure and pull the request again.

Issue #413, opened 2020-04-13
[Spell] Summon Creature (not read char location)


alexdan2:
When cast a summon creature NPC not placed corrected under char feet but save every last spell location.

How reproduce:
- Cast a poison wall (s_poison_field)
- Cast summon creature (s_summon_creature)
- Summoned npc is placed on location targetted by field not under char feet
- If not previus cast make when login get a server error (on save) because summoned NPC not have any location:
- Other spell like s_summon_elem_air work correct
```
12:58:ERROR:GC: Deleted UID=01fee, Defname='c_rabbit', Name='rabbit'. Invalid code=0x3202 (Object not placed in the world).
```

Extra info:
- Tested on clear save and clear comunity script pack
- Tested on 3262 Build

JulianUO:
Can you confirm this has been fixed on the latest commit?

alexdan2:
Just tested on Build 3273, and work perfect. Thx !

PR #420, opened 2020-04-20
Several ports from 56d.

JulianUO:
I'm realizing that the @SkillPreStart fix might be useless as skill is set in the argument of the Skill_OnCharTrigger, but it should not break anything.


JulianUO:
Included a fix for issue #413, I didn't want to mess with Spell_TargCheck() but maybe the real m_Act_p should be set there?

Issue #432, opened 2020-04-26
Last Commit create a bug with archery


Jhobean:
How to reproduce
1- Equip an i_bow and i_arrow
2- Attack someone
3- During you attack hit do .stone
4- You continu attacking but you CANT initiate a new attack.

EDIT: Same bug with Dagger

Jhobean:
A work!  thx

Issue #409, opened 2020-04-11
Lost target after hiding/invisibility and cannot attack him anymore


pentanorl:
When a player under attack become hidden or invisible, the attacker lose the target and cannot attack him anymore until he targets someone else, only then he can attack his previous target again.
I checked the attacker list and the "non-selectable" player remains into the list.
Any idea what's causing this?

Here's a video I found on Discord about this same issue.
https://www.dropbox.com/s/4kgfr4xu6xpv2qk/Hid.mp4?dl=0

xwerswoodx:
Seems like you are not using default client, have you tried it on default client?

xwerswoodx:
I confirm this issue.

alexdan2:
Me too, make some test and happen same. 
Tested on different build (new, and old one 10d ago) same.

Massi17:
same issue with paralyze

 https://github.com/Sphereserver/Source-X/issues/342

pentanorl:
any news about a possible cause?

FiLoUs64:
We have this problem too. Can be very OP in PVP

pentanorl:
OP and annoying, ‘cause if you are 1vs1 there is no way to overcome this bug, and when you lose the possibility to target your opponent, be sure that you are going to die badly. ^^

fjgo86:
Sent a commit for this, this will affect NPCs too, so any feedback covering the hiding issue against both players and npcs would be appretiated.

GitMadKing:
@fjgo86 build failing.

xwerswoodx:
I tried it with latest build and issue still continue, when I tried to attack with X char to Y it starts and when Y goes invis X stop and when Y continue attack to X after reveal, X doesn't attack the char at all while it should call an attack when get hit, but the attack totally stops for X until peace/war.

Jhobean:
I confirm fix do not work.

alexdan2:
I make some test and i can also confirm still bugged

xwerswoodx:
I fixed the issue and send a pull request but it can take time to accept, so I put here the fixed version, so you can try and tell me if there is still issue.

https://www.mediafire.com/file/1uls29llqyri55b/SphereSvrX64_nightly.rar/file

alexdan2:
@xwerswoodx hello, i tested your file and work good :) thx

Jhobean:
It's not work for me in the last build

alexdan2:
@Jhobean 
[...]

Issue #397, opened 2020-04-02
@NPCRestock - Color of item not correct SET.


alexdan2:
Hello, 
When call @NPCRestock for add item on NPC body lilke:
`ON=@NPCRestock
    ITEM=i_shirt_plain
    COLOR=colors_all
`
Happen this problem:
- Correct add i_shirt_plain on NPC body.
- Wrong set color, Color is set on NPC Body not on ITEM.

Thx.
Alessandro

GitMadKing:
item that npc wear is under @create and not @npcrestock
what do really need it for under @npcrestock
?

alexdan2:
Hello, i use for backwards compatibility, but i can move all @create trigger.
i try code but happen this:

https://ibb.co/F69GB1c

cbnolok:
Npcrestock has its sense as a trigger for that stuff, btw can you copy the script on pastebin? I suspect, as GitMadKing said, that there's a newline error

alexdan2:
Hello i check for new line and make new clear npc whitout events ecc and happen same issue.
Here pastebin of script: https://pastebin.com/SmtnSgWT
Here a ingame video: https://www.youtube.com/watch?v=cyHfomQodAc&feature=youtu.be

I also see when npc spawn color is OK, after 0.5/1 sec he take color or weapon.
other information i can share is take color of last item in resotck, npc take 33 color.
ES:
```
	ITEM=i_dagger
	COLOR=2
	ITEM=i_apron_half
	COLOR=33
```

Thx

GitMadKing:
on notepad++ activate "reversed P"  
an post screenshoot. (you can simply paste the screen and it.will host kn github)

alexdan2:
This is my test script:
![image](https://user-images.githubusercontent.com/6176986/78339575-fa1c7500-7594-11ea-9126-5bdc936e1390.png)

GitMadKing:
is the result here
![image](https://user-images.githubusercontent.com/16067766/78339858-731bcc80-7595-11ea-831f-3385a7362cc0.png)

maybe a duplicate char?

try ths:
```
[CHARDEF C_HWB]
NAME=#NAMES_HUMANMALE
ID=C_MAN
DESIRES=i_gold,e_notoriety,t_log,t_ingot
AVERSIONS=t_TRAP
ON=@Create
    TITLE=l'Armaiolo
    NPC=brain_vendor
    COLOR=55
    STR={86 100}
    DEX={86 100}
[...]

Issue #358, opened 2020-01-19
Emotecolor


orishi:
Please, if already not included, add <emotecolor> property to characters, as they did on 56d.
https://github.com/Sphereserver/Source/commit/5b29e68c2b58694c86bc181f62e9926606d0d2c7

meaowR:
would be great cause i want to change some emotes to green or yellow too :)

orishi:
I'm afraid you didn't understand what I needed.
I need to send SAYU in the same color as the client/player has set for himself. I need to read client's emote speech color settings in client options.

As on 56d:
https://github.com/Sphereserver/Source/commit/5b29e68c2b58694c86bc181f62e9926606d0d2c7

xwerswoodx:
Firstly, the code you send is a color override for emotes. So I will do same thing but in different way because both versions has different codes.

Secondly, we already have SpeechColorOverride for SAY overrides, so could you check it? Because it's already exist. If it's not working for SAYU, I can check.

orishi:
Well, maybe it's just that english is not my native language, so let me rephrase.

First of all, this is already implemented and will display "some text" in the same color as the client/player has set in the client.exe options:
`sayua <speechcolor>,0,0,1 some text`

I need the same thing but for emote color. Or did I not understand your code and it's solved?

xwerswoodx:
Hmm, I guess I get you know, I implemented another EMOTECOLOR setting to get emote color from client-side. I will return after I made some test.

orishi:
Great we understand each other now, let me know, thanks.

orishi:
Works great, thanks.

PR #435, opened 2020-04-26
Fix Update

xwerswoodx:
- Added: EmoteColor to players to get Client-Side emote color of the player. (Issue #358)
	Note:	To set emote colors for players, the players use emote speech at once.
		If the player change his emote color, he need to use emote ": message" to sphere override his default emote color.
		Also, make sure Player has no EmoteColorOverride on himself to make sure emote color change for player.
- Fixed: Stoned, paralyzed, sleep characters continue attacking. (Issue #409)
- Fixed: Can't attack again after the characters stoned. (Issue #432)
- Possible Fix: COLOR variable change the color of human if the human is shop keeper (vendor). (Issue #397)
- Fixed: Unreferenced variables for new GMPage system to avoid possible issues.
- Fixed: Archery bug occurs when the attacker is stoned. (Issue #432)
- Fixed: Paralyze causes the attacking issue. (Issue #342)


xwerswoodx:
Because of the conflict script, seems like I made a ";" mistake while cleaning unnecessary lines, and I can't fix it anymore...

Jhobean:
Include invul flag too?

xwerswoodx:
Yes, when the target is goes invul (which players shouldn't be invul) it cancels the target, so after invul/stone you can still continue attack with dclicking the target. I haven't make it to wait for attack like invis/hidden because stone and invis is unexpected behaviour for players.

Issue #441, opened 2020-04-27
xmfhtmlgumpcolor not work correct


alexdan2:
In last release bug happen in xmfhtmlgumpcolor (possibile other function).

**Result in Build 3277:**
![image](https://user-images.githubusercontent.com/6176986/80428774-f73e4700-88ea-11ea-802d-3a85be92e7e1.png)

**Result in Build 3285:**
![image](https://user-images.githubusercontent.com/6176986/80428791-01f8dc00-88eb-11ea-8442-07c87f5243ee.png)

**Example of code affected:**

```
[DEFNAME DIALOG_COLORS]
dcolor_white					16777215

xmfhtmlgumpcolor 10 13 530 20 <ddef.scp.craft_title_<ctag0.craft.skill>> 0 0 <ddef.dcolor_white> // menu header````

```
**Simple code for test (call whit .dialog d_TESTTEST) :**
```
[dialog d_TESTTEST]
40,40

page 0
resizepic 0 0 9200 530 437   // background/border
gumppictiled 10 10 510 417 2624   // darker background
gumppictiled 10 32 510 5 5058   // 1 grey separators - start here
gumppictiled 10 287 510 5 5058   // 2
gumppictiled 10 337 510 5 5058   // 3
gumppictiled 210 37 5 250 5058   // 4
gumppictiled 160 292 5 45 5058   // 5 grey separators - end here
checkertrans 10 10 510 417   // transparency layer
xmfhtmlgumpcolor 10 13 530 20 1044001 0 0 16777215 // --> this part affected by bug

```

xwerswoodx:
I checked the latest commits, but haven't seen any changes about it. Have you changed your client version? Maybe it's because of Client packets.

alexdan2:
I remake test whit simple code:

```
[dialog d_TESTTEST]
40,40

page 0
resizepic 0 0 9200 530 437   // background/border
gumppictiled 10 10 510 417 2624   // darker background
gumppictiled 10 32 510 5 5058   // 1 grey separators - start here
gumppictiled 10 287 510 5 5058   // 2
gumppictiled 10 337 510 5 5058   // 3
gumppictiled 210 37 5 250 5058   // 4
gumppictiled 160 292 5 45 5058   // 5 grey separators - end here
checkertrans 10 10 510 417   // transparency layer
xmfhtmlgumpcolor 10 13 530 20 1044001 0 0 16777215 // --> this part affected by bug```

**Result in Build 3277:**
[...]

Issue #440, opened 2020-04-27
Trigger @NPCRestock still make item


alexdan2:
When fire @NPCRestock still add item to NPC backpack and not check if exist or not.

**How Reproduce**
- Add a spawn of a vendor like c_butcher
- Set Region tag.RestockVendors=10 for fast restock (if not want use this tag have to wait 10 mins)
- Open vendor backpack and see.

**Additional info:**
- Tested on last build (3285)
- Clear script pack
- Happen only on **Brain_vendor**
- Possible connected to issue #397 fix

Little video clip: https://www.youtube.com/watch?v=_jexZ9i18d8

**NPC Script code:**
```
[CHARDEF c_butcher]
ID=c_man
NAME=#NAMES_HUMANMALE the butcher
DESIRES=i_gold
AVERSIONS=t_trap
TSPEECH=spk_human_default
TSPEECH=spk_shopkeep
CATEGORY=Civilized
SUBSECTION=Tradesmen
DESCRIPTION=Butcher (human male)
ON=@Create
   COLOR=colors_skin
   NPC=brain_vendor
   NEED=i_gold
   STR={50 75}
   DEX={45 55}
   INT={45 55}

   ANATOMY={20.0 30.0}
   ANIMALLORE={50.0 75.0}
   COOKING={20.0 30.0}
   FENCING={50.0 75.0}
   FISHING={20.0 30.0}
   FORENSICS={20.0 30.0}
   HERDING={20.0 30.0}
   MAGICRESISTANCE={20.0 30.0}
   SWORDSMANSHIP={20.0 30.0}
   TACTICS={20.0 30.0}
   TASTEID={20.0 30.0}
   WRESTLING={20.0 30.0}

   ITEMNEWBIE=random_light
   ITEMNEWBIE=random_male_hair
   COLOR=colors_hair
   ITEMNEWBIE=random_facial_hair,1,R5
   COLOR=match_hair
   ITEM=random_shirts_human
   COLOR=colors_all
   ITEM=random_pants_human
   COLOR=colors_all
   ITEM=random_footwear_human
   COLOR=colors_neutral
   ITEM=i_apron_half
   COLOR=colors_brown
ON=@NPCRestock
   ITEM=random_coin_purse
   ITEM=i_cleaver_meat,1,R3
   ITEM=i_knife_skinning,1,R3
   ITEM=i_knife_butcher,1,R3

   SELL=VENDOR_S_BUTCHER_TEMPLATE
[...]

Issue #439, opened 2020-04-27
Impoosible attack someone after he had invul flag


Jhobean:
Seem to be like https://github.com/Sphereserver/Source-X/issues/409

The last fix did not fix it.

How reproduce:
1- Take 2 character
2- Both hit each other
3- One become invul
4- After remove the invul flag, it's impossible to attack him

It's ok with the stone flag.

There was a partil fix in this COMMIT: https://github.com/Sphereserver/Source-X/pull/435

xwerswoodx:
Hm, I remember I checked invul flag but if I am not I can send a quick fix for that with my next pull request.

xwerswoodx:
Hmm I checked the code and I remember correct that I checked invul flag, but now I find out another issue that is little confusing, because when you are invul and not hit anyone after that others can attack you when you invul off. But if you attack someone, he can't attack you anymore as he war/tab or dclicking. This seems like when invul person attack to another person, server directly save that person as invul for other player and it doesn't reset until server restart.

xwerswoodx:
Hmm seems like this is not an issue at all, the reason is when someone tried to attack to someone invul, sphere add him to ignore list to avoid unnecessary continuous attack search, so it takes time to remove ignore status.

Jhobean:
Ok so in pvm I cannot use invul flag???  It was working on 56d. Dit not try in 55i

xwerswoodx:
Hmm, maybe I can do that ignore thing as pvp only. Or you can RETURN 1 under @GetHit event instead of giving INVUL.

Jhobean:
There a way to force "actualise" ignore list?

xwerswoodx:
Yeah there is;
```
27-01-2015, XuN
- Updated @CombatDelete trigger with a new argn1 value, now it's telling Sphere to remove or keep 1/0 the attacker's memories when removing him from the list.
**- Added @HitIgnore trigger firing when one character is going to avoid an ignored target in the target selection: argn1=1 keep him ignored, argn2= un-ignore target (so character can attack him again).**
- Added @Eat trigger, read wiki for more info: http://wiki.sphere.torfo.org/index.php/@Eat
```

Jhobean:
Wooo. Ok. So I put  @CombatDelete on each player event.  On this trigger i put
 if (src.defname==mymonster)
Return 1
Endif

xwerswoodx:
I guess
```
On=@HitIgnore
[...]

Issue #436, opened 2020-04-26
Add script keywords REMOVECLILOC and REPLACECLILOC.


cbnolok:
I think those keywords would be really useful, so that people could modify/remove the hardcoded clilocs, instead of having to softcode them all.

alexdan2:
Yeah is a very nice feature!

xwerswoodx:
Hm, I can inform that, I added that feature to source, i will pull the request asap.

PR #443, opened 2020-04-28
Some Fix and Improvements

xwerswoodx:
- Possible fix for issue #440 and #397 without change @NPCRestock behaviour.
- Added: TOPCONT as reference top container item. (It returns item's itself if it's equipped or on ground)
	The difference between TOPCONT and TOPOBJ is TOPCONT only returns the top container not characters.
	If you use topCont on item inside bankbox or inside the subcontainers of bankbox, it returns bankbox uid.
- Fixed: Improved Julian's bank check code for works on the items inside sub containers. (Issue #429)
- Added: ARGN2 under @Attack trigger to replace Ignore status. (Issue #439)
	ARGN2 = 1 //Force to ignore attacker.
	ARGN2 = 0 //Force to unignore attacker.
- Added: @ClientTooltip_Sent trigger which triggers after default tooltips sent. (Issue #436)
- Added: REMOVECLILOC and REPLACECLILOC functions to remove/replace default tooltip properties under @ClientTooltip_AfterDefault trigger. (Issue #436)
- Added: CLILOCLIST on objects to get list of the clilocs.
- Added: CLILOCCOUNT and CLILOC.n[.ID|.VAL] to get clilocs.
- Added: Added missing gumps and gump default sizes.
- Improvement: Wrong calculation for height for Containers but still can't override Ultima Online itself maximum X, Y locations. (Issue #423)
- Fixed: Mass Curse doesn't lower the stats. (Issue #446)
- Fixed: Fixed the issue on XMFHTMLGUMP, XMFHTMLGUMPCOLOR and XMFHTMLTOK. (Issue #441)
- Fixed: Not cleaning skill when @PreStart trigger return as true. (Issue #450)


src/game/clients/CClientMsg_AOSTooltip.cpp
```diff
@@ -112,6 +112,13 @@ bool CClient::addAOSTooltip(CObjBase * pObj, bool fRequested, bool fShop)
 
                 pObj->AddPropsTooltipData(pObj);
 			}
+			
+			if (IsTrigUsed(TRIGGER_CLIENTTOOLTIP_SENT) || (pItem && IsTrigUsed(TRIGGER_ITEMCLIENTTOOLTIP_SENT)) || (pChar && IsTrigUsed(TRIGGER_CHARCLIENTTOOLTIP_SENT)))
+			{
+				CScriptTriggerArgs args(pObj);
+				args.m_iN1 = fRequested;
+				iRet = pObj->OnTrigger("@ClientTooltip_Sent", this->GetChar(), &args); //Save to return on iRet to make sure return value doesn't stuck the boolean.
```
cbnolok, line 120: Tip: it's faster to use the enum instead of the string

src/game/items/CItem.cpp
```diff
@@ -49,6 +49,7 @@ lpctstr const CItem::sm_szTrigName[ITRIG_QTY+1] =	// static
 	"@Buy",
 	"@Click",
 	"@ClientTooltip",	// Sending tooltip to a client
+	"@ClientTooltip_Sent",
```
cbnolok, line 52: Out of curiosity, what are the use cases of this new trigger?

src/game/CObjBase.cpp
```diff
@@ -1971,6 +2017,59 @@ bool CObjBase::r_Verb( CScript & s, CTextConsole * pSrc ) // Execute command fro
...
+                
+                for (size_t i = 0; i < m_TooltipData.size(); i++)
+                {
+                	CClientTooltip* ct = m_TooltipData[i].get();
+                	if (ct->m_clilocid == clilocid)
+                	{
+		                if ( g_Cfg.m_iDebugFlags & DEBUGF_SCRIPTS )
+        	    	        g_Log.EventDebug("SCRIPT: replacecliloc(%u,'%s')\n", clilocid, sLocArgs.GetPtr());
+        	    	    m_TooltipData.erase(m_TooltipData.begin() + i);
+        	    	    m_TooltipData.emplace(m_TooltipData.begin() + i, new CClientTooltip(clilocid, sLocArgs.GetPtr()));
```
cbnolok, line 2067: Another tip: vector's emplace method is very nice also because you can avoid manually creating the object, you can directly pass the constructor's arguments like this:
`m_TooltipData.emplace(m_TooltipData.begin() + i, clilocid, sLocArgs.GetPtr());`
This results in shorter code and lets the compiler use the best way to create the object (which isn't by manually using `new` since in this case we are using an unique_ptr, it should be with std::make_unique)

src/game/items/CItem.cpp
xwerswoodx: The reason to add new trigger is before remove or replace the cliloc, it should be send by sphere so we need another trigger which trigger after default sphere clilocs to use REPLACECLILOC and REMOVECLILOC. Because @ClientTooltip is trigger before default sphere tooltips send so when you use replacecliloc it doesn't see default clilocs because it doesn't send yet. I didn't want to move @ClientTooltip under default cliloc because RETURN 1 not working anymore to block default cliloc, so I add another trigger which triggered after default cliloc sent.

src/game/clients/CClientMsg_AOSTooltip.cpp
xwerswoodx: Yeah, but I just copied ClientTooltip trigger and it uses string, so I haven't changed the way, because I haven't sure about if someone doing it for some reason, but I can change it to ITRIG enum.

src/game/CObjBase.cpp
xwerswoodx: I haven't known about it, I am updating the code quickly.

src/game/clients/CClientMsg_AOSTooltip.cpp
xwerswoodx: Hmm seems like CScriptObj only support lpctstr because when I change it to ITRIG_CLIENTTOOLTIP_SENT it gives that error

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	'TRIGRET_TYPE CScriptObj::OnTrigger(lpctstr,CTextConsole *,CScriptTriggerArgs *)': cannot convert argument 1 from 'ITRIG_TYPE' to 'lpctstr'	spheresvr	C:\Users\user\Documents\GitHub\Source-X\src\game\clients\CClientMsg_AOSTooltip.cpp	120
```

src/game/CObjBase.cpp
xwerswoodx: When I changed emplace with that way vs gives that error;
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	'std::unique_ptr<CClientTooltip,std::default_delete<CClientTooltip>>::unique_ptr(const std::unique_ptr<CClientTooltip,std::default_delete<CClientTooltip>> &)': cannot convert argument 1 from 'dword' to 'CClientTooltip *' (compiling source file C:\Users\user\Documents\GitHub\Source-X\src\game\CObjBase.cpp)	spheresvr	C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\include\xmemory	671	
```

cbnolok: You're right, std::unique_ptr constructor doesn't accept an argument list, so you should directly pass a `std::make_unique<CClientTooltip>(clilocid, sLocArgs.GetPtr())`

src/game/items/CItem.cpp
cbnolok: Got it! What about renaming ClientTooltip_Sent to something more self-explanatory like ClientTooltipDefault? ClientTooltip_Sent makes me think that it triggers after the tooltip list is sent to the client, not that we are after default tooltips are generated but before they are sent

xwerswoodx: Basically where I put the trigger is meaning nearly sent, I mean;

The query of the process is;
- Start function
- Call @ClientTooltip trigger
- Sent default aos cliloc list
- Call @ClientTooltip_Sent trigger
- Finish function

This is why I chose add _SENT, we can't use @ClientTooltipDefault, because when you use RETURN 1 under @ClientTooltip, default tooltips doesn't send but @ClientTooltip_Sent still trigger to replace or remove clilocs which set under @ClientTooltip. We can maybe use @ClientTooltip_AfterDefault or something like this but it is longer and hard to remember, but we can use anything specific to make sure it's understandable

src/game/CObjBase.cpp
xwerswoodx: I replaced the code with std::make_unique

src/game/clients/CClientMsg_AOSTooltip.cpp
cbnolok: You're right, there are no better alternatives here

src/game/items/CItem.cpp
cbnolok: Checking the code, to me it looks that the following is happening, correct me if i'm wrong:
- Start function
- Call @ClientTooltip trigger 
- If there's no RETURN 1, add to the tooltip cache the default tooltips
- Call @ClientTooltip_Sent

xwerswoodx: Yes, I forgot to specify for hash calculate, yes this is the place where ClientTooltip_Sent triggered.

xwerswoodx: Maybe we can change the trigger as @ClilocSent or @ClientTooltip_Cliloc(Sent)

cbnolok: Still i consider a "sent tooltip" a tooltip for which the packets have been sent to the client, but that's actually happening after ClientTooltip_Sent, the only thing happening between ClientTooltip and ClientTooltip_Sent is the default tooltips being added to the list, so i can't really understand why you are using the "sent" word to name that tooltip, to me it makes more sense to give a name that recalls to the fact that the default tooltips were just generated

xwerswoodx: Honestly I haven't thing about it that much, I just want to make similar with ClientTooltip, and I thought clilocs already sent before that trigger so I just make is @ClientTooltip_Sent. But I can change it to @ClientTooltip_AfterCliloc or @ClientTooltip_ClilocSent to make more understandable. I haven't think about tooltip is a packet, I just think it's a trigger which triggers after clilocs sent, so I just add _Sent at the end.

xwerswoodx: I replace the trigger with @ClientTooltip_AfterDefault to be more specific so people can understand easily that can use add/replace/removecliloc under @ClientTooltip_AfterDefault which trigger after default clilocs sent by server.

Issue #442, opened 2020-04-28
Wrong notoflag when two guildstones are in war.


Massi17:
Create two guildstones, enlist a character in each guild, declare war and you will see your opponent with the notoflag-ally (green)

pentanorl:
Looks like something is broken with notoriety colors related to guild status.
Just create different guildstones without declaring war nor alliances and you'll see everyone ORANGE (Enemy) instead of their natural notoriety colors (Blu/Red/Gray). The moment you declare war from both sides you'll see your enemy GREEN (Ally).
Of course since you'll see everyone guilded as green or orange, kills and criminal actions don't trigger at all.
Guildmates results GREEN (Ally) as supposed to, so looks like that just War/Peace colors are inverted and peace doesn't has to be considered as alliance (when guild aligns are different).


Already tested with latest build and some previous with community pack.
This is the untouched *.ini stuff related to notoriety colors
```
// Default color settings for characters names (notoriety)
//ColorNotoGood=059			// blue
//ColorNotoGuildSame=03f	// green (ally)
//ColorNotoNeutral=03b2		// grey (can be attacked)
//ColorNotoCriminal=03b2	// grey (criminal)
//ColorNotoGuildWar=090		// orange (enemy guild)
//ColorNotoEvil=022			// red
//ColorNotoInvul=035		// yellow
//ColorNotoInvulGameMaster=0b	// purple
//ColorNotoDefault=03b2		// grey (if not any other)
```

cbnolok:
As a note to other devs and contributors, we should evaluate if it would be
useful to port the 56d guild and town system (and the chat one?) since they
were greatly reworked but i don't know if any functionality was changed

Il mar 28 apr 2020, 13:12 Stefano <notifications@github.com> ha scritto:

pentanorl:
I just made this simple function to help debug.
The function prints out the list of notovalues showing the respective color for each player inside the list.
And well, something strange happen.
You should set the "NotoTimeOut" (*.ini) to a low value (just a few seconds) to speed things up.

**This is one scenario:**
I have three characters everybody with positive karma to simplify the situation.
Two of them are in the same guild [AAA] and a third one is in a guild at war [BBB]:
- Player One [AAA] //(myself)
- Player Two [AAA] //(guildmate)
- Player Three [BBB] //(enemy)

[...]

Issue #369, opened 2020-02-23
PacketDeath


alexrmrm:
PacketDeath make anim of "you are death" but not activate black screen

xwerswoodx:
Have you tried to turn on sphere.ini, PacketDeathAnimation setting?

// Enable death screen animation on classic clients (on enhanced clients this is always enabled)
PacketDeathAnimation=1

alexrmrm:
yes i try make anim and show msg of you are death but not black screen.

xwerswoodx:
Hmm, I don't know why that happened, I am trying and packet is sending for me, which version of uo are you using? If you are using Razor or Orion they can block packets or render packets different.

alexrmrm:
client 203 without any.

criminalx:
https://puu.sh/FFnp2/24e52470ce.mp4
that account was gm, but it happen same with player account.

sphere.ini:
// Enable death screen animation on classic clients (on enhanced clients this is always enabled)
PacketDeathAnimation=1

criminalx:
Also you can't pass through a door while dead

GitMadKing:
i can confirm, client 3.0.6j
even if logout and login back, dead, still see in color

criminalx:
It is not fixed yet

alexrmrm:
Not work that fix. continue same problem.

xwerswoodx:
Hmm, I tried with my latest build and I got you are dead animation when I died (tested on 2.0.3 client)

xwerswoodx:
Just for make sure, do you have any @Death trigger? Maybe you can use @Death to send player some safe area when he is dead. If you return 1, he is technically don't die so animation doesn't work. Even nothing done about this from all these time but as I said before I had no issue from the beginning, and after 5 fixes, if the issue haven't solved, I personally think it's some script side issue. Maybe you need to check all of your @Death @Kill @DeathCorpse triggers.

cbnolok:
Is this still an issue?

PR #468, opened 2020-05-10
Issues #369, #442 and other minor.

JulianUO:
This bug was ported from 56 source. I'm having a small issue with resurrect on older clients (where the client is crashing). And my SpyUO isn't working right now to check what is making it crash, will investigate on it later if someone can reproduce it.


JulianUO:
I miss pulled, sorry.

PR #476, opened 2020-05-15
Fixes for #369, #442, minor issue and new keywords/refs.

JulianUO:
Changelog extended:
- Added: Multi property "MULTIOFFSET x,y,z", it contains the offset for displaying the target multi (as in multi creation). This way, you can create dynamic targets with multi, for example, using a pentagram as cannon target.
- Added: Ship keywords ANCHOR and DIRFACE, will return the anchor status (0/1) and the direction the ship is facing (0 to 7). They are not writable, use the ship commands for that.
- Fixed: House type wasn't saved with the world save, when loading the server a guild or public house was loaded as a private houses.
- Fixed: Stones (guilds) not checking correctly for ally/war members, added a check for the member privs for ally/war since a enemy is just another member with specific privs in the stone.
- Fixed: Older clients do not support frontfall in death packet, should research on which client this is enabled and add a CanSendTo function. (Issue #369)


src/network/send.cpp
```diff
@@ -3506,11 +3509,12 @@ PacketMessageUNICODE::PacketMessageUNICODE(const CClient* target, const nword* p
  ***************************************************************************/
 PacketDeath::PacketDeath(CChar* dead, CItemCorpse* corpse, bool fFrontFall) : PacketSend(XCMD_CharDeath, 13, PRI_NORMAL)
 {
+	UNREFERENCED_PARAMETER(fFrontFall);
```
cbnolok, line 3512: What about sending fFrontFall if the client version supports it, instead of sending 0 to everyone?

JulianUO: I'm investigating which client mincliver supports it, but to make Sphere usable to everyone I did this modification for now.

cbnolok: Okay then, merging!

Issue #496, opened 2020-07-23
Adding scroll to spellbook doesn't add the good one


FiLoUs64:
There a problem with scrolls when you add it to spellbook. For example with blade spirit when you add it it appear as Clumsy.

Also we have problems where spells in fourth and fifth circle disappear for no reason.

See video : https://cdn.discordapp.com/attachments/476635430821953567/735682030679097375/Ultima_Online_-_Phil_Cryptonite_2020-07-22_22-15-22.mp4

Scroll script : 

> [ITEMDEF 01f4d]
DEFNAME=i_scroll_blade_spirits
NAME=Parchemin%s% d'esprit de lame
TYPE=t_scroll
VALUE=27
WEIGHT=0.1
RESOURCES=1 i_scroll_blank,1 i_reag_black_pearl,1 i_reag_mandrake_root,1 i_reag_nightshade
SKILLMAKE=Inscription 50.0,Magery 40.0
CATEGORY=Provisions - Alchemy & Magic
SUBSECTION=Scrolls - 5th Circle
DESCRIPTION=Blade Spirits Scroll
ON=@Create
   ATTR=attr_magic
   MOREX=s_blade_spirits
   MOREY={50 80}

FiLoUs64:
Fixed. Thanks!

PR #497, opened 2020-07-23
Fix for issue #496: Adding the Blade Spirits spell to a spellbook added the Clumsy spell.

drk84:
This also caused a corruption of the more1 value of the spellbooks, probably causing the disappearance of the rest of the spells?
They will have to replace players spellbooks i guess


cbnolok:
Doesn't this cause another bug? It reverts that, suboptimal at this point, commit

https://github.com/Sphereserver/Source-X/commit/892048b5e39e1df94bdca90a59a3c9d92aeff6ee

drk84:
I tried adding the 32nd and 64th spell (recall and summon water elemental) and i am not getting the bug, but but to be sure what if i replace the checks:
**if (i < 32u)** and  ** if (i <= 64u)**
with
**ifuint(spell) <= 32u)**and ** if (uint(spell) <= 64u)**
?

cbnolok:
If that works 😁 i have no pc on this week so i can't help with testing

drk84:
Ah true i can't just use the
` if (uint(spell) <= 32u) and if (uint(spell) <= 64u)` 
because spells from other spellbooks  have an higher index.

Issue #499, opened 2020-07-31
Cant create areas in the new maps (ilshenar, ter mur, etc)


RodVIII:
AREADEF do not work on new maps. I just tested with the latest sphere 56d and it works fine in there.

[AREADEF a_TheTradingPost]
NAME=The Trading Post
P=1107,405,-80,2
RECT=1088,384,1125,422,2
GROUP=Trading Post
FLAGS=region_flag_nobuilding|region_antimagic_gate|region_antimagic_recall_out|region_flag_insta_logout|region_flag_safe

cbnolok:
Do you mean that it works only for map 0? Does it create an area on the wrong map (0 instead of 2) or doesn't it create it at all?

RodVIII:
Yes. I can only make areas on map0. Maybe map1 too, but not on the rest.

Jhobean:
Seem to work.  Look picture 
Left before fix, Right After
![image](https://user-images.githubusercontent.com/51728381/93693079-1079a700-fac9-11ea-828a-6b173c6e0d2e.png)


```
[AREADEF a_twisted_weald]
NAME=Twisted Weald
```

PR #501, opened 2020-08-28
Possible Fix for issue #499 : Cant create areas in the new maps (ilshenar, ter mur, etc)

drk84:
The AREADEF were loaded correctly, even the new ones, infact you could access the various property by using serv.areadef.defname.xx

It seems that the problem was caused in the method CSectorList::Init(), there the variable:

`int iSectorIndex = 0;`

was declared outside the following FOR cycle

`for (int iMap = 0; iMap < MAP_SUPPORTED_QTY; ++iMap)`

Inside this cycle there is another FOR that initialize every sector of the map Sphere is currently looping

`for (; iSectorIndex < iSectorQty; ++iSectorIndex)`

After this cycle ends, Sphere will loop through the next map and try again to initialize its sectors but because iSectorIndex was not reset to 0 at the end of the map loop, the sector cycle will fail because iSectorIndex would have already the value of iSectorQty from the previous map cycle.


drk84:
In pratically from what i understand the AREADEF were loaded but weren't linked properly to the relative sectors.
So in someway Sphere was always returning the topmost region in map1,2,3,4,5 etc

PR #505, opened 2020-09-09
Osx build + travis osx build script

Nirad:


src/common/sphere_library/CSAssoc.h
```diff
@@ -11,7 +11,7 @@
 
 
 #ifndef OFFSETOF			// stddef.h ?
-	#define OFFSETOF(s,m)   	(int)( (byte *)&(((s *)0)->m) - (byte *)0 )
+	#define OFFSETOF(TYPE, ELEMENT) ((size_t)&(((TYPE *)0)->ELEMENT))
```
cbnolok, line 14: Did you have the chance to test this change on 32 and 64 bits build, and on platforms different than Mac OS?

src/common/sphereversion.h
```diff
@@ -46,6 +46,9 @@
 	#else
 		#define SPHERE_VER_FILEOS_STR		"[FreeBSD-32]"
 	#endif
+#elif __APPLE__
+    #define SPHERE_VER_FILEOS                0x0L    // VOS_UNKNOWN
```
cbnolok, line 50: Can you consider the possibility of a 32 bits build on Mac?

src/game/clients/CClientLog.cpp
```diff
@@ -222,8 +222,8 @@ bool CClient::addRelay( const CServerDef * pServ )
 		CSString sCustomerID(pServ->GetName());
 		sCustomerID.Add(GetAccount()->GetName());
 
-		dwCustomerId = z_crc32(0L, Z_NULL, 0);
-		dwCustomerId = z_crc32(dwCustomerId, reinterpret_cast<const z_Bytef *>(sCustomerID.GetBuffer()), (z_uInt)sCustomerID.GetLength());
+		dwCustomerId = crc32(0L, Z_NULL, 0);
```
cbnolok, line 225: Was it necessary to remove the z_ prefix? Afair, we have a zlib config header with the Z_PREFIX macro set to 1, so i can't think why that's necessary

travis.sh
```diff
@@ -0,0 +1,19 @@
...
+echo TRAVIS_BUILD_DIR is $TRAVIS_BUILD_DIR
+echo TRAVIS_OS_NAME is $TRAVIS_OS_NAME
+
+if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    export HOMEBREW_NO_AUTO_UPDATE=1
+    brew install ninja
+    brew install mysql-client
+    gcc -v && g++ -v && cmake --version
+    echo Building Release
+    mkdir -p release && cd release && cmake -G Ninja ../src -DCMAKE_BUILD_TYPE=Release $EXTRA && cmake --build . || exit 1
```
cbnolok, line 15: We had tons of problems with Travis-CI trying to upload stuff to the ftp server, or even doing apt-get updates with a sudo-enabled environment. Can you make a similar section in the AppVeyor script?
Also, can you switch from a Release to an automated Nightly build and split the build commands so that we have only one command per line? Imho we have better output in case of an error.
Thanks!

src/common/sphereversion.h
Nirad: https://www.infoq.com/news/2018/01/macos-32bit-app-deprecation/ 32 bit build is Deprecate by Apple

src/game/clients/CClientLog.cpp
Nirad: https://github.com/Sphereserver/Source-X/blob/master/src/cmake/toolchains/Linux-GNU-64.cmake#L83 yes i have see it here, but for some reason the compiler dont get it let me check that again

src/common/sphere_library/CSAssoc.h
Nirad: i have tested it on Windows 64 bit versions let me do the test for 32 bit one. I will get back to you

src/common/sphereversion.h
cbnolok: It's deprecated, but it's still possible if someone is motivated enough, right? To cover every possible case i'd like to add also 32 bits and even arm arch

Issue #506, opened 2020-09-20
Invisible GMs don't see other invisible GMs


orishi:
1) Invisible GMs don't see other invisible GMs - basically same plevel should always see each other, that's a bug on X

2) I would love to make it so that all plevels higher than 1 see each other while invisible. Could you add this option to sphere.ini?

Jhobean:
Here the Pull Request: https://github.com/Sphereserver/Source-X/pull/522

orishi:
Perfect, works as I need it. Thanks.

orishi:
When you use CanSeeSamePLevel=4 and Player with plevel 1 logs out of game, he can still be seen by GMs. However when player with plevel 4 or higher logs out, he disappears just fine.

When setting this ini to 0, everything works like it used to.

Jhobean:
Do you have allshow active?

Jhobean:
Yes there other bug too...  2 plevel 0 can see each other when they are hide:smile: i'll fix this soon

Jhobean:
I push a Fix for the bug with Plevel 0.  @orishi ABout your Plevel1 bug, I'm not able reproduce it. If you see loged out player it's because you have allshow active or because you have linger time on logout.

orishi:
It was ClientLinger, I didn't realize I have it set up for 300 instead of 0. It's not used for GM accounts.

PR #507, opened 2020-09-21
Fix for GM don't see same PLEVEL GM hidden (issue #506 )

Jhobean:
Like explained in issue #506 2 GM with same Plevel are not able to see each other when they are invisible.

On the script, the result of comparaison of plevel need to have a "0" to be able to see the target. The comparaison <= was not appropriate because same Plevel give a "1".


Nirad:
👍

cbnolok:
I'd like to add that as an optional feature, like a new ini setting: CanSeeSamePLevel. I'm open to other ideas though!

GitMadKing:
agree. not as default. but optional YES


---- Attivato dom, 18 ott 2020 19:04:02 +0200 cbnolok <notifications@github.com> ha scritto ----




I'd like to add that as an optional feature, like a new ini setting: CanSeeSamePLevel. I'm open to other ideas though!

Jhobean:
Ini setting like this?
//Gm with same level can see each other = 0
// Gm with same plevel dont see each other = -1
// All plevel over x see all gm = x (2 to 7)
CanSeeSamePLevel=0

Jhobean:
I dont really understand the purpose of this, only wish to fix the actual bug:smile:

FiLoUs64:
> I dont really understand the purpose of this, only wish to fix the actual bug😄

They said it's not a bug... was like this in 55i I think, you can't see same plevel

cbnolok:
Yes like that, simply because they changed the default behavior since 56, it's not a bug

GitMadKing:
add a (55 like) with -1does sphere has a default value for unset ini settings?i mean, if after this update, i change the "exe" but not the sphere.ini is a default in core to some value and then "updated" by sphere.ini settings or what?Marius ---- On mer, 21 ott 2020 06:54:45 +0200  notifications@github.com  wrote ----
Yes like that, simply because they changed the default behavior since 56, it's not a bug

Jhobean:
Replace By this one https://github.com/Sphereserver/Source-X/pull/522

Issue #463, opened 2020-05-08
REVEALF_OSILIKEPERSONALSPACE = Walking on someone Hide show message


Jhobean:
EDIT: First issue was not an issue. Just read message down

GitMadKing:
Did you looked here?
**sphere.ini**
```
// Reveal flags related to Invisible Spell
// REVEALF_DETECTINGHIDDEN		0001	// Reveal Spell with Detecting Hidden Skill.
// REVEALF_LOOTINGSELF			0002	// Reveal when looting self bodies (NOTE: Hidden characters will be always revealed by default).
// REVEALF_LOOTINGOTHERS		0004	// Reveal when looting bodies of other Players or NPCs (NOTE: Hidden characters will be always revealed by default).
// REVEALF_SPEAK			0008	// Reveal when speaking.
// REVEALF_SPELLCAST			0010	// Reveal when starting to cast a Spell.
// REVEALF_OSILIKEPERSONALSPACE		0020	// Do not reveal when a character enters on personal space (and do not send any message).
// REVEALF_SNOOPING			0040	// Do not reveal hidden players while snooping.
// REVEALF_STEALING			0080	// Do not reveal hidden players while stealing.
// REVEALF_STEALING_SUCCESS		0100	// Reveal hidden players when stealing successfully finished.
// REVEALF_STEALING_FAIL		0200	// Reveal hidden players when stealing failed.
// Set to 04 to make it behave like 55i.
```

ACTIVATE THIS **REVEALF_OSILIKEPERSONALSPACE 020**

it works, but has a problem, get the message you steped on a person,
**change name on the issue to : REVEALF_OSILIKEPERSONALSPACE message of walking trough**

![hidden_personal_space](https://user-images.githubusercontent.com/16067766/81475427-7cd7c600-920c-11ea-9331-eec8455c8736.gif)

Jhobean:
You are right!  The ini work but the problem is the message appear:
![image](https://user-images.githubusercontent.com/51728381/81481678-c1109980-91ff-11ea-9689-d61c557384a2.png)

THX

JulianUO:
Use
ON=@PersonalSpace
return 0

PR #509, opened 2020-09-25
FIX personnal space message when invisible (Issue #463)

Jhobean:
Like it's explain on this issue: https://github.com/Sphereserver/Source-X/issues/463.

When Walking on someone invisible and Setting is set, your not suppose to have a message:
`// REVEALF_OSILIKEPERSONALSPACE		0020	// Do not reveal when a character enters on personal space (and do not send any message).`

For now, you have a message.

I just added a restriction if the player is invisible it's not show message. IF REVEALF_OSILIKEPERSONALSPACE is not activated, the character is already reveal at this line.


cbnolok:
Can't you use RETURN 0 in @PersonalSpace instead?

Jhobean:
Hmmm. Yes it can be script side. But on the ini file it's write there no message if you are invisible. It's why I think it's abug.

cbnolok:
It's not a bug, since that may be a desired behavior and you can use scripts to change it. Though you are correct on the ini, i'll change the description!

PR #513, opened 2020-10-04
Fix for commit #510.Typo error.

Jhobean:
WHen I pushed the commit #510 , I made a mistake and did not save my LAST modification. Newbie error. :-)

Here with this modification, all working well. Was not working because you was checking the Plevel of the npc.


cbnolok:
It was my fault actually 😅

Issue #514, opened 2020-10-07
Item stacking ini setting glitch the dupelist


Jhobean:
When activating this option: 
`// EF_ItemStacking				00000004 // Enable item stacking feature when drop items on ground`
![image](https://user-images.githubusercontent.com/51728381/95279127-e0980680-081f-11eb-8d88-1451b9c4ca9c.png)

Item stack when we drop them on the ground. But when this option is activate. there no way to change the dupelist on an item. For exemple i_dagger. Normaly the item switch side.

I think the "bug" is HERE 
https://github.com/Sphereserver/Source-X/blob/892048b5e39e1df94bdca90a59a3c9d92aeff6ee/src/game/chars/CCharAct.cpp#L2033

There is a return few line before. The purpose seem to evitate to flip item. I think it should be flip if its the first item of the pile.

And I see an other bug too.. `( g_Cfg.m_fFlipDroppedItems || pItem->Can(CAN_I_FLIP) ` I think it should be 
`( g_Cfg.m_fFlipDroppedItems && pItem->Can(CAN_I_FLIP) `  Because the setting in the ini is useless like this 

```
// Flip dropped items
FlipDroppedItems=1
```

PR #515, opened 2020-10-07
Fix Item do not flip when EF_ItemStacking is active (Issue  #514)

Jhobean:
This commit fix 2 issues relating to flip item.
1- When EF_ItemStacking is active it was impossible to flip an item on drop
2- FlipDroppedItems on ini was useless. Active or inactive there was no difference

In the same time, I ported from 56d the SphereCrypt.ini


Jhobean:
Fix this issue https://github.com/Sphereserver/Source-X/issues/514

PR #522, opened 2020-10-21
Fix for GM don't see same PLEVEL GM hidden + feature (issue #506 ) 

Jhobean:
Like explained in issue #506 2 GM with same Plevel are not able to see each other when they are invisible.

With this modification, we can see it


Jhobean:
Don't merge now it do not work

orishi:
Can you please add option that every plevel over 3 (3 or higher) can see each other?

GitMadKing:
just set the ini setting to 3



// Possibility to see Hidden GM 



// GM with same plevel can see each other = 0

// GM with same plevel dont see each other = 1

// All plevel over 2 see all each other = 2

CanSeeSamePLevel=0





---- Attivato mer, 21 ott 2020 20:04:49 +0200 Vojtěch Šťastný <mailto:notifications@github.com> ha scritto ----




Can you please add option that every plevel over 3 (3 or higher) can see each other?

Jhobean:
@orishi with option 2 all plevel over 4 can see other. I choose 4 because by defaut .invis available at plevel 4. Mayne I van change to 3 but why 3?  Why not 2?:confused:

orishi:
> @orishi with option 2 all plevel over 4 can see other. I choose 4 because by defaut .invis available at plevel 4. Maybe I can change to 3 but why 3? Why not 2?😕

Well, my .invis command is set to plevel 3. It would be better if you could choose the plevel and not hard code it to 4 or higher.

Jhobean:
Since you made the feature request. I put Plevel 3 and ajust the .ini description.

Jhobean:
@cbnolok I think I have a mistake on my code. Because I put Return true, it will exit the function "Cansee"? If yes, i'll be able to see A disconnect GM without have .allshow active.

Jhobean:
Here you go it's work well Without Return true

cbnolok:
Sorry for bringing this in now, but what about having this:
0 and 1 as you coded, then 2-7 to set the plevel starting from you are able to see everyone, independently to their plevel?

Jhobean:
Sound fair i'll try

Issue #534, opened 2020-11-11
When OF_PetSlots	 is active, I can tame an animal when maxfollower already reach


Jhobean:
Fixed on 56 with this commit:
https://github.com/Sphereserver/Source/commit/ba905fec97974b3ea0c0db7bc94e120b92c7cd0f

PR #536, opened 2020-11-12
AnimalTamingnot checking follower slots (Issue #534) 

Jhobean:
Effective only when OF_PetSlots is activate

Fix Issue #534


drk84:
Confirmed

Issue #456, opened 2020-05-03
Little bug in vendor sell menu


alexdan2:
When i set custom price in vendor buy list, sell menu not correct read PRICE but use standard item VALUE, but when i complese sell he give correct gold.

**Template Code:**
```
[TEMPLATE VENDOR_B_ALCHEMIST]
CATEGORY=Vendor Templates
SUBSECTION=Buy Templates
DESCRIPTION=Alchemist
BUY=i_potion_strength
PRICE=10
```

This is wrong:
![image](https://user-images.githubusercontent.com/6176986/80928044-2862bf80-8da2-11ea-8e18-c7874b135b87.png)

This is correct:
![image](https://user-images.githubusercontent.com/6176986/80928052-43353400-8da2-11ea-9995-04c0644bee00.png)

xwerswoodx:
what do you mean with custom price?

GitMadKing:
was not to use buyvalue sellvalue
?

alexdan2:
I not know this buy and sell value can link me docs pls. Is possible this fix this issue than set Direct in buylist
--
Inviato da Libero Mail per Android Martedì, 05 Maggio 2020, 08:35PM +02:00 da Mad King  notifications@github.com :

Jhobean:
I think the value must be in the itemdef of the item.. Never use the value on the template. it's suppose to work?

alexdan2:
I use value in template for set different sell price.. 
For example if want npc sell red potion to 800gp but not want alchemy player can sell to high ratio value automatic calculated from 800gp i set PRICE=10 on vendor template.

NPC give correct PRICE=10 but show different in sell gump.

alexdan2:
@GitMadKing 
I try to find SELLVALUE usefull guide but nothing on wiki, also i try to put SELLVALUE=10 on item but not work also on template.

Jhobean:
The bug seem to be there
Where the packet is send to the client, the server use the itemdef value instead of the item value.
Don't know how to fix it.
https://github.com/Sphereserver/Source-X/blob/4b73c9c73efb1a2235297743059ff11fec6eb1ab/src/network/send.cpp#L2266

Jhobean:
hmmm  Don't seem to have a bug. 
![image](https://user-images.githubusercontent.com/51728381/94294614-125ec280-ff2e-11ea-80fd-69d0b576b790.png)

The price is ajusted.... Did you restock your npc?

alexdan2:
[...]

PR #545, opened 2020-11-24
Fix sell system do not show/give correct gold (Issue #456)

Jhobean:
This commit Fix 2 issues:
- Issue #456 (price not correct on dialog when price is define on template)
EX: ```
BUY=i_tongs,{4 18}
price=5000
``
- When you have an item with VALUE=10,20 the gold give to you was not depending on the quality of the item (because it was looking the itemdef item with quality =0)

Before this fix, Item sold to NPC was showing correct value on the display:
image
But when we accept the trade we received always the cash for an item with quality of 0

here what we do now:
1- Always priorize the value/price set on template
2- if there no price/value set on template, always use the value/price of the item itself (in packpack of player when selling  and itemdef if on vendor.


Jhobean:
- I removed all the possibility to check the "PRICE" when selling an item. On some script error, it was possible to player to set "price" with their vendor NPC.

- I fixed the dialog when selling an item and it's now always show the real money you will receive

- Added tag.override.value can be use to ajust the value of an item because value is R only.

I don't think modify Buy system is require.

Issue #525, opened 2020-10-31
coloring for hidden players


meaowR:
if i change colors
```
ColorInvis=04001            // 04001 = transparent color, 0 = default
ColorInvisSpell=04001        // 04001 = transparent color, 0 = default (This one is for s_invisibility spell, this includes the invis potion.)
ColorHidden=04001          // 04001 = transparent color, 0 = default
```
in .ini it looks like coloring my body only :slight_smile:
![map](https://user-images.githubusercontent.com/61579076/97791546-7f780e80-1be4-11eb-82ab-f13aeda869da.jpg)

Jhobean:
Made some test, bug occur on these 2:
ColorInvisSpell=04001
ColorHidden=04001

The first one is correct

Jhobean:
Ok some more info about this bug.

- Paperdoll skin or item should never change.
- Item on the anim should get the Hue selected in the ini setting.

Jhobean:
Fix Merge Need feedback

PR #547, opened 2020-11-27
FIX: When hiding/invis item color was not change (issue #525)

Jhobean:
There was 2 bugs

- When we hide/invis [Spell or potion], the item on the paperdoll was not taking the grey HUE
- When we modify the color of invis in the ini setting, the anim character AND the paperdoll was not changing.

image

This is a Fix for the issue #525 and a fix for the paperdoll


cbnolok:
As far as i remember, on Sphere none of the paperdoll items were ever colored when invis/hidden, am i wrong? In that case, maybe an ini switch may be added to turn on the recoloring of the whole paperdoll, since it would be a feature change and not a fix, strictly speaking

Jhobean:
What about The color of the skin?

Jhobean:
Ok we must choose what we want about the paperdoll: 
1- Stay like now (only skin color change) (except insubtantial you have your paperdoll change too)

2- Like OSI (nothing change on paperdoll)
image

3- Like ServUo (All paperdoll change)

cbnolok:
Historically sphere behaved like osi i think, without altering the skin on
the paperdoll.
I'd only add a switch to enable paperdoll coloring with the color of the
colorinvis being used

Il mar 1 dic 2020, 03:04 Jhobean <notifications@github.com> ha scritto:

Jhobean:
Ok after verification, 55i and OSI is the same.
There is no modification on the paperdoll at all. I'll Try to make it work like this

Jhobean:
With the last commit, I setted back the old 55i  and OSI like situation.

- When you hide, invis yourself, your skin color do not change on the paperdoll (Since 56d, your skin color was changing)

About the feature where you can set different color, for hidding or invis in the ini:
- Your paperdoll color AND your anim take the color you set.
It's an obligation because the Grey color on the anim is done by the client and not the server. If you want override this grey color, you must manually set the skin and item color (affect paperdoll as well)

- For doing it work, we must send a Full packet when we get hide and when we use invis spell. Avoid send a full packet was doing the cloth color did not change.

PR #549, opened 2020-12-02
Change the hue of invis Item just before send packet to client.

Jhobean:
The setting ColorInvisItem change the Hue of invis item and permit the GM to make difference between invis item or not.

The last method had a bug and the global gethue fonction was use to change the color of item. It was causing some problem and sometime the real COLOR of the item was taking the "fake" hue.

With this commit you get the "fake" hue just before send the packet to the client.

### DIFF
--- a/src/game/items/CItem.cpp
+++ b/src/game/items/CItem.cpp
@@ -1901,7 +1901,7 @@ bool CItem::SetName( lpctstr pszName )
 	return SetNamePool( pszName );
 }
 
-HUE_TYPE CItem::GetHue_visualforclient() const
+HUE_TYPE CItem::GetHueVisible() const
 {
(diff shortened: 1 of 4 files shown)


src/game/items/CItem.cpp
```diff
@@ -1901,7 +1901,7 @@ bool CItem::SetName( lpctstr pszName )
 	return SetNamePool( pszName );
 }
 
-HUE_TYPE CItem::GetHue() const  //Override of CObjBase::GetHue()
+HUE_TYPE CItem::GetHue_visualforclient() const
```
cbnolok, line 1904: Can you change the method name with a shorter and less "dispersive" one? Like GetHueVisible (which also respects the naming convention with the uppercase letters).

cbnolok (changes requested):

Issue #528, opened 2020-11-05
RANGE - for bows and crossbows


golfincz:
RANGEH and RANGEL are poorly displayed for bows and crossbows. RangeH shows 1 and RangeL 0.

Jhobean:
The bug occur there: https://github.com/Sphereserver/Source-X/blob/a04e0329f452802be25ef77f804ad9c6ba0f507f/src/game/components/CCPropsItemWeapon.cpp#L74

_uiRange is always 0

Jhobean:
I added THis
https://github.com/Jhobean/Source-X/blob/5c63d4e7a4771fe58aad8222943ae8da4c9e0287/src/game/components/CCPropsItemWeapon.cpp#L97-L101

and it's work...  I don't know how. I just copied the range setting 10 line under..

Maybe @fjgo86, @cbnolok or @coruja747  may have the explanation.

cbnolok:
As far as i am concerned, it looks right to me, why should that be a problem?

cbnolok:
Oh well i misread, of course before reading that value, you are setting range via scripts, is that right?

golfincz:
Yes according to the script. And especially firearms need to have a range that is different for crossbows and different for bows (It is very individual for each weapon). And the scripted own combat system then calculates the damage, among other things, according to the extent of the weapon.

xwerswoodx:
I can agree that bug happened to me after i updated my sphere to latest one, but i haven't updated it for a while so i have no idea which version I used last. I tried the rangeh and l before updating and they had no problem after i updated latest version and it starts return high as 1 and 0 as low

Jhobean:
Ok I confirm the bug appear here https://github.com/Sphereserver/Source-X/commit/a04e0329f452802be25ef77f804ad9c6ba0f507f?branch=a04e0329f452802be25ef77f804ad9c6ba0f507f&diff=split

PR #551, opened 2020-12-02
Fix where the range L and range H did not give the correct value. (Issue #528)

Jhobean:
This commit is for fixing the issue #528 .

When the core want to get the value of RangeL ou RANGEH with this call:
https://github.com/Sphereserver/Source-X/blob/0fbe39fd7c54e7a4367c8c6018784acdd8774bce/src/game/CEntityProps.cpp#L237

The value _uiRange was always empty.

I added a return false if _uiRange is empty. So the core continue his process and try to get the range later when _uiRange have now a value

https://github.com/Sphereserver/Source-X/blob/0fbe39fd7c54e7a4367c8c6018784acdd8774bce/src/game/CEntityProps.cpp#L284

I not undestand at 100% why _uiRange is empty at the beginning.


cbnolok:
_uiRange is empty if not set via scripts, but, even in that case, the method should return a default value (which is 1 even for melee weapons if i'm not mistaken). That's the reason why it didn't return false if the range wasn't set.

Jhobean:
Ok the bug was here since this PR https://github.com/Sphereserver/Source-X/pull/479.
Like you said Nolok, it's require to send 1 value if _uiRange ==0 because normal weapon need a range of 1

Issue #502, opened 2020-09-08
Fixweight Fonction do not work


Jhobean:
When I do .fixweight I have a BUG:
![image](https://user-images.githubusercontent.com/51728381/92534574-22bf2f80-f203-11ea-9173-aec58a12235c.png)

My total weight drop to 6 and when I drop a item to the grounnd I go in negative . The only Way I found to fix it is reboot the server.

I use debug and put a stop point on this function:
https://github.com/Sphereserver/Source-X/blob/f59514b38b1ec25c3ee3e121d3db6000cf4a3ea1/src/game/CContainer.cpp#L94

The return is always 0

Issue #487, opened 2020-05-25
Kills and Deaths Counts


canavar2212:
hello I want to add multiple kills and deaths count system on the server. for example, I want to separate it as KILLS, KILLSNPC and KILLSPLAYER. KILLSPLAYER is running, but the reason I added KILLSNPC NOT WORKING WHAT HAPPENS?

Murder Kills;
`
case CPC_KILLS:
sVal.FormatVal( m_wMurders );
return true;
`
`case CPC_KILLS:
			m_wMurders = (word)(s.GetArgVal());
			pChar->NotoSave_Update();
			return true;
`
in r_WriteChar

`if ( m_wMurders )
		s.WriteKeyVal("KILLS", m_wMurders);
`

in CworldImport 
IMPFLAGS_CHARS
`
			else if ( s.IsKey("KILLS" ) && pChar->m_pPlayer )
			{
				pChar->m_pPlayer->m_wMurders = (word)(atoi(pArg));
			}
`

in CcharPlayer_Props.tbl

`
ADD(KILLS,			"KILLS")
`

in CcharPlayer
```
	m_wDeaths = m_wMurders = 0;
	m_wDeathsNPC = m_wMurdersNPC = 0;
	m_wDeathsPlayer = m_wMurdersPlayer = 0;
```

it works when I add the same for KILLSPLAYER but not when I add it for KILLSNPC.

canavar2212:
No problem when I do the same for deaths.

DEATHS, DEATHSPLAYER, DEATHSNPC

the problem only occurs when adding KILLSNPC.

console output ;

`22:36:ERROR:Undefined keyword 'killsnpc'`

xwerswoodx:
In CCharPlayer_Props.tbl you need to make sure the it's sorted in alphabetically, if not it doesn't see the command.

canavar2212:
Evet, alfabetik sıraya göre yapınca sorun çözüldü. Teşekkür ederim.

Issue #485, opened 2020-05-23
Spells: Poison - monsters stop completly attacking


razyhost:
hello,

when i do cast poison on a monster(easily tested on dragon) and trigger a stealth/hide ability (or run away), wait a few seconds, unhide, the monster doesn't attack me anymore, i can freely cast on it without being attacked, but anyone other approaching is still attacked by the monster automatically as it should

Step to reproduce: i suggest on a dragon.

-> Cast Poison
-> wait the spell land
-> stealth (or simply run away) 
-> wait at least one tick of poison (a 3-4sec)
-> reveal yourself (come back if you did it the running way!)
-> Cast any offensive spell, harm/magic arrow

At this stade the dragon won't attack you anymore for around 10min. you can even tame it that way.

If any player other than me come near that dragon, the dragon will attack him on side and still ignore me.

drk84:
Can't reproduce it :\

Jhobean:
Ok Here the way to reproduce it (with a Gm it's more easy but the bug is the same with regular player)
![image](https://user-images.githubusercontent.com/51728381/93691941-6135d380-faba-11ea-96f4-eb153519f044.png)

1- .go despise
2- .add c_elemental_earth
3- .invul OFF
4- Wait the red message golem attack you
5- .cast 20  on the golem
6- Run from his LOS (go downstair, hide, run etc)
7- Wait 5 sec
8- Come back and walk next to him like you walk near a rock­.

Massi17:
it doesn't seem to happen if you use the invisibility spell, but it does happen with the hiding skill as explained above.
1 - mob attacking you
2 - player hide
3 - mob going home
4 - player got revealed

Jhobean:
Suppose to be fixe. Need to be test

alexdan2:
I testet whit this way 10 / 15 time and now sems work good, npc back to attack when come back in LoS.

> 1- .go despise
> 2- .add c_elemental_earth
> 3- .invul OFF
> 4- Wait the red message golem attack you
> 5- .cast 20 on the golem
> 6- Run from his LOS (go downstair, hide, run etc)
> 7- Wait 5 sec
> 8- Come back and walk next to him like you walk near a rock­.

PR #557, opened 2020-12-05
Fix for NPCs stop attacking after being poisoned NPC (Issue #487) and Fixweight not working properly (Issue #502).

drk84:
-Fixed: NPCs stop attacking after being poisoned NPC (Issue #487).

The bug happens when a NPC get damaged while the player is not reachable and it was caused by a missing assignement in the check in method CCharFight::OnTakeDamage

```
 if (bAttackerExists == false)
  {
   ....
   attacker.ignore = false; //Attacking.ignore was not set causing it to be set at 246
   m_lastAttackers.emplace_back(std::move(attacker));
  }
```

When the player goes away, the player is properly removed from the NPC's attacker list,  but if in the meantime the NPC suffers damage from the player by a command/item that
calls directly OnTakeDamage the NPC will add back the player but without setting the attacker.ignore property properly (It was set to 246 in my case).

-Fixed: Fixweight not working properly (Issue #502).

The problem was caused by the  method call:

```
 m_totalweight += pCont->GetWeight();
...
int CItemContainer::GetWeight(word amount) const
{ // true weight == container item + contents.
 return( CItem::GetWeight(amount) + CContainer::GetTotalWeight());
}
```

CContainer::GetTotalWeight would return m_totalweight of the current container but at the top of the Fixweight method we set m_totalweight value to 0 with the result
that Fixweight returned only the weight of the single item containers.

Minor problem i found:
When a player login, it seems that the Bank box weight it's added to is own weight, so a naked player should only weights 3 stones (backpack) but instead it weights 6 stones (3 backpack + 3 bank box)
I can't manage to track it.

@drk84
Forgot to remove the //

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -2319,3 +2319,8 @@ Must have event, even if it's empty, since it's applied by the source to generat
 		change the gold you will receive when selling it to a NPC. PRICE is now exclusive to player vendor.
 - Added: tag.override.value can now be add to an item to change his value when trading with a NPC. In scripts, on vendor template 
 		(BUY= or SELL=). For changing the value of the itemdef, you must use tag.override.value=xxx instead of old syntax price=xxx.
+		
+		
+05-12-2020, Drk84
+-Fixed: NPCs stop attacking after being poisoned NPC  (Issue #487).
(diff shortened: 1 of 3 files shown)


Jhobean:
I think it's the issue https://github.com/Sphereserver/Source-X/issues/485 #485 Not 487

Issue #68, opened 2018-05-06
[PORT]72ff09966548b27829fcf2ed7608207ae6862d26, Fixed: Resurrect, Reveal, Meteor Swarm and Lightning spells showing effect animation


fjgo86:
To be tested more and eventually ported.
Revision: https://github.com/Sphereserver/Source/commit/72ff09966548b27829fcf2ed7608207ae6862d26
Author: Coruja coruja@outlook.com
Date: 12/11/2016 20:41:49
Message:
Fixed: Resurrect, Reveal, Meteor Swarm and Lightning spells showing effect animation even when EFFECT_ID=0 is set
[sphereCrypt.ini]: Added crypt key for classic clients 7.0.54 ~ 7.0.55 and enhanced clients 4.0.54 ~ 4.0.55

Issue #59, opened 2018-05-06
[PORT]dd7a59b451e9d4d8d99c3bec15b330cf60dded2f, Fixed: Items inside trade window not firing @DropOn_Item trigger when the trade move the item to player bacpkack


fjgo86:
To be discussed and eventually ported.
Revision: https://github.com/Sphereserver/Source/commit/dd7a59b451e9d4d8d99c3bec15b330cf60dded2f
Author: Coruja coruja@outlook.com
Date: 02/08/2016 07:07:12
Message:
Fixed: Items inside trade window not firing @DropOn_Item trigger when the trade move the item to player bacpkack

Issue #58, opened 2018-05-06
[PORT]f26b39b9b40fba76660c36d57baa8387d3613272, Fixed: Security issue setting on account login


fjgo86:
Revision: https://github.com/Sphereserver/Source/commit/f26b39b9b40fba76660c36d57baa8387d3613272
Author: Coruja coruja@outlook.com
Date: 22/07/2016 22:52:03
Message:
Fixed: Security issue setting account login as chat name when newest clients try to setup the chat window for the first time and the char name is not available
Fixed: Invisible chars being incorrectly revealed by REVEALF_LOOTINGOTHERS reveal flag when picking items from the ground
Fixed: Return 0/1 on spell triggers not working correctly
Fixed: Function MOVENEAR not working correctly

PR #565, opened 2020-12-11
[Port 56d] 3 Commits from 56d

Jhobean:
**SphereX Issue #59**
Fixed: Items inside trade window not firing @DropOn_Item trigger
I tested and the issue was there. The fix by Coruja is now well apply on X.

**SphereX Issue #68** 
12-11-2016, Coruja
Fixed: Resurrect, Reveal, Meteor Swarm and Lightning spells showing effect animation even when EFFECT_ID=0 is set

Resurrect: Fixed like 56d
Reveal: Work differently on X no need of this
Light: Added the effect because it was wiped on X
Meteor Swarm :  On X it seem to be removed

**SphereX Issue #58** 
22-07-2016, Coruja
Fixed: Security issue setting account login as chat name when newest clients try to setup the chat window for the first time and the char name is not available.

I only removed the possibility to have account name as chat name. Description for the Fix was insane

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -2320,7 +2320,11 @@ Must have event, even if it's empty, since it's applied by the source to generat
 - Added: tag.override.value can now be add to an item to change his value when trading with a NPC. In scripts, on vendor template 
 		(BUY= or SELL=). For changing the value of the itemdef, you must use tag.override.value=xxx instead of old syntax price=xxx.
 		
-		
 05-12-2020, Drk84
 -Fixed: NPCs stop attacking after being poisoned NPC  (Issue #487).
 -Fixed: Fixweight not working properly (Issue #502).
+
+11-12-2020, Jhobean
+- [Port from 0.56d, 01-08-2016, Coruja]
+	Fixed: Items inside trade window not firing @DropOn_Item trigger when the trade move the item to player bacpkack.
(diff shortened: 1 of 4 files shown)


cbnolok:
Can you explain more about meteor swarm and lightning? At a quick look it seems to me a legit fix

Jhobean:
About the  meteor swarm, it's the same thing than lot of spell. EFFECT_ID=  change nothing on the script. On X, all spell seem to be rework and effect in script don't have interraction.
For exemple, if I set 
```
[SPELL 55]
DEFNAME=s_meteor_swarm
EFFECT_ID=i_fx_fireball
```
The meteor will send a fireball on the char and smoke will be on the char..  It's work very well.. But if I do:
`EFFECT_ID=0`
It happen the same thing.  So effect_ID do nothing on lot of spell.

About Lighting it was a mistake, lighting send a lighting on the head and it's work fine. The modif was for s_light. I added it because X did not have effect for this

Jhobean:
About the  meteor swarm, it's the same thing than lot of spell. EFFECT_ID=  Effect is already well do. On X, lot of spelll seem to be rework and effect in script are well implement interraction. If you set 0 or any effect it work well.

I have difficulty to make good test with script pack because it seem to have some problem with spell configuration. When I test with my script it work well. Bug in script pack will be find later.

About Lighting it was a mistake, lighting send a lighting on the head and it's work fine. (if you add an effect_id it's add this effect over the lighting)  
The modif of coruja was for s_light. I added it on the commit because on X it was impossible to have an effect for this spell. Now, I think it's fine admin can add effect via script.

Issue #64, opened 2018-05-06
[PORT]c8223f203bbeb753ff0e1d3388db13e4df0b4d4a, Fixed: Internal check to prevent drop items inside walls preventing the item drop even when the wall is on another floor


fjgo86:
To be tested more and eventually port.
Revision: https://github.com/Sphereserver/Source/commit/c8223f203bbeb753ff0e1d3388db13e4df0b4d4a
Author: Coruja coruja@outlook.com
Date: 21/08/2016 05:29:13
Message:
Fixed: Internal check to prevent drop items inside walls preventing the item drop even when the wall is on another floor
Fixed: Message "Too many items here!" not showing correctly when items are dropped on areas with too many items

Jhobean:
I confirm this line is not working: `iDecayTime = 60 * TICK_PER_SEC;		// force decay (even when REGION_FLAG_NODECAY is set)`

We have some player making potion and drop them to ground at the same spot. We have like 2000 item on ground

Jhobean:
The 2 bug don't occur on X. Exception about the wall where GM can place item on a wall so the item go up to the top level floor.

because it,s a GM, I think it's correct

```
if (pt.IsValidPoint())
    {
        if (pCharMover && pCharMover->IsPriv(PRIV_GM))
		    ptNewPlace = pt;
        else
        {
            const CPointMap ptNear = CWorldMap::FindItemTypeNearby(pt, IT_WALL, 0, true, true);
            if (!ptNear.IsValidPoint())
                ptNewPlace = pt;
        }
    }
```

cbnolok:
So it isn't happening when gm mode is off, right?

Jhobean:
Yes, I confirm with GM off item go back in packback if you try to drop on wall.  We could close this issue

PR #567, opened 2020-12-14
FIX: Too many item message and add possibility to change it

Jhobean:
When droping an item to the ground and there was too many item, the message was display on the character journal instead of ON the item. It's because the update was not done and your client did not see the item. By putting an Update before, it's fix.

THis part fix the issue like coruja did on #64.

I added the possibility to customize the message by script. For exemple, a french shard don't want to see "Too many items here!"

Don't forget approuve the new script pack message: https://github.com/cbnolok/SphereX-Scriptpack/pull/60


Nirad:
Do you need to add this mesage to the other script pack as well to ?

cbnolok:
As a suggestion, some comments with old code are left there for a reason, it's better to preserve it instead of wiping it

Jhobean:
@Nirad i have pr old of 1 year on other script pack (56d)   I dont do PR there anymore

src/game/items/CItem.cpp
```diff
@@ -1561,29 +1561,31 @@ bool CItem::MoveToCheck( const CPointMap & pt, CChar * pCharMover )
...
-			if ( iItemCount > g_Cfg.m_iMaxItemComplexity )
-			{
-				Speak("Too many items here!");
-				iDecayTime = 60 * MSECS_PER_SEC;		// force decay (even when REGION_FLAG_NODECAY is set)
-				break;
-			}
+		++iItemCount;
+		if ( iItemCount > g_Cfg.m_iMaxItemComplexity )
+		{
+			Speak(g_Cfg.GetDefaultMsg(DEFMSG_TOO_MANY_ITEM));
```
cbnolok, line 1582: Can you rename it to DEFMSG_TOO_MANY_ITEMS please?

cbnolok (changes requested):

Issue #359, opened 2020-01-20
Tooltips not showing


orishi:
Confirmed on Discord discussion by drk84, X is not showing tooltips for client <6.0.50. I'm currently using 5.0.9.0. On 56 version the tooltips work without any troubles.

> drk84 today at 13:37
> no no
> confirmed
> version before 6.0.50
> does not show up
> the tooltip

meaowR:
check your ini or scripts...all fine for vanilla X

orishi:
> check your ini or scripts...all fine for vanilla X

We already did a lot of testing, it's not the scripts or ini. Confirmed with drk84 on discord, as I already stated above, that this is a bug and that the tooltips are only showing for clients >6.0.50

orishi:
Any progress on this? Sorry to bump that, but we kinda need it :(

meaowR:
> Any progress on this? Sorry to bump that, but we kinda need it :(

no
anyway i decided to not using tooltips anymore )))

orishi:
I'm still stuck on 56 on this "project", too bad. Unfortunately I cannot use higher client version without making huge changes to it :(

Nirad:
@orishi is it posible to ask you where i can find your mul file?

orishi:
@Nirad you don't need muls to replicate this, just download basic ML client with 5.0.9.0 patch

Nirad:
@orishi you can test the core with this fix here : https://ci.appveyor.com/project/Nirad/source-experimental-c49x7/build/job/n4lhisy36m0ny0i6/artifacts

orishi:
@Nirad For character it works fine. It doesn't work for items, they don't show any tooltip at all.

Jhobean:
??? 
![image](https://user-images.githubusercontent.com/51728381/102224029-7355c100-3eb3-11eb-891c-18ed749229e6.png)

orishi:
It's probably something in my scripts then. Lets wait if there is more feedback from other users. I'll check it later again.

Jhobean:
> There something strange. At your first connexion it's do not work because you dont have the Version configured. Log out log in and it's work.

Ok "Bug" found. First login your tooltip is not activate because your account dont have clientversion on the save. Just reboot the client and it will be fine. After the next save your tooltip will work for ever.  I confirm this Fix work on my side

orishi:
It was script error on my side, it's working correctly.

Jhobean:
Its not merge yet

PR #568, opened 2020-12-15
In older client, client version is sent to late after characterslection.

Nirad:
to fix this issue #359  we send a ClientVersionReq after server list the we can setup the charlist and client features flags properly.

56D code:
https://github.com/Sphereserver/Source/blob/3d3c141fed72bb09bbdc4a1369fc0a410962ff15/src/graysvr/CClientLog.cpp#L649-L654
https://github.com/Sphereserver/Source/blob/191a13e7bc91a709db79d1270c467a5561038541/src/network/receive.cpp#L2653-L2655


Nirad:
@Jhobean it's fix it on my side but if you can test with one of your other older version.

Jhobean:
I tested with original UO clients 5.0.9.1, 6.0.0.0 and it's work know 
image

This bug was for sure create other bug with old client version.

Issue #554, opened 2020-12-03
TIMERF STOP, Seems not work


alexdan2:
I make some test for use STOP on timerf but call STOP not do nothing.

I call .f_test_Start for start the timer whit f_test_say.
When i call .f_test_Stop whit STOP on f_test_say, timerf still present

This is code i use for test:
```
[FUNCTION f_test_Start]
SRC.SYSMESSAGE Start
SRC.TIMERF 100, f_test_say
IF (<SRC.ISTIMERF.f_test_say>)
	SRC.SYSMESSAGE Present
ELSE
	SRC.SYSMESSAGE NOT Present
ENDIF
RETURN 1

[FUNCTION f_test_Stop]
SRC.TIMERF STOP, f_test_say
IF (<SRC.ISTIMERF.f_test_say>)
	SRC.SYSMESSAGE Present
ELSE
	SRC.SYSMESSAGE NOT Present
ENDIF
RETURN 1

[FUNCTION f_test_say]
	SRC.SYSMESSAGE EndTIME
RETURN 1
```

Jhobean:
Remove the space between , and fonction name

alexdan2:
Oky whitout space work :( damn syntax

Jhobean:
@alexdan2 i know it because this already cause lot of headhache on my side!!:grin:

cbnolok:
Actually it's worth fixing

Nirad:
so just trim space shold work

Nirad:
REF. https://github.com/Sphereserver/Source-X/blob/master/src/game/CObjBase.cpp#L2575-L2608

alexdan2:
Thx just tested, work good whit or whitout space.
Thx

Issue #531, opened 2020-11-07
TAGAT functions for account tags


orishi:
Request to add functions for account.tag similar to character/item tags.

TAGAT.index | R | Gets a TAG at the given zero-based index.
TAGAT.index.KEY | R | Gets the name of the TAG at the given zero-based index.
TAGAT.index.VAL | R | Gets the value of the TAG at the given zero-based index.
TAGLIST | W | Outputs a list of the object's TAGs.

xwerswoodx:
Just for note, TAGLIST already exists for accounts

alexdan2:
Tested and work good.

I used this way for test and perfect:

```
[FUNCTION f_add_account_tag]
SRC.ACCOUNT.TAG.HELLO1 = 1
SRC.ACCOUNT.TAG.HELLO2 = 1
SRC.ACCOUNT.TAG.HELP = 1

[FUNCTION f_add_account_tag_list]
SERV.LOG Account TAGCOUNT: <SRC.ACCOUNT.TAGCOUNT>
FOR 0 <EVAL <SRC.ACCOUNT.TAGCOUNT> -1>
	SERV.LOG TAG NAME: <SRC.ACCOUNT.TAGAT.<DLOCAL._FOR>.KEY> 
	SERV.LOG TAG VALUE: <SRC.ACCOUNT.TAGAT.<DLOCAL._FOR>.VAL> 
	SERV.LOG ----------------
ENDFOR
```

Issue #512, opened 2020-09-30
Creating item with ID= use always the itemdef dispid


Jhobean:
When creating an item, the core always use DISPID of the main item and not the dispid we asked.
SO the ID= dont take the dispid

It can be fix by script but it's lot of job, all multi must be ajusted.

EX:  .add m_fountain_brown
![image](https://user-images.githubusercontent.com/51728381/94689092-6ede2b00-02fc-11eb-8c56-c20a0bcb66be.png)

NORMALY ID= all spec of this itemdef (weight, etc) and must exclude the dispid and take the dispid you writed

```
[ITEMDEF m_fountain_brown]
ID=019d3
NAME=Fountain
TYPE=t_multi
CATEGORY=Multis
SUBSECTION=Misc
DESCRIPTION=Fountain (Brown)
MULTIREGION=0,0,0,0
COMPONENT=019cf,0,1,0
COMPONENT=019ce,0,2,0
COMPONENT=i_fountain_2,0,3,0
...
```

Second Exemple:
If I create a new item
```
[ITEMDEF i_cocoon_mephitis1]
DEFNAME=i_cocoon_mephitis1
ID=010da //(dupelist item of i_web_cocooon)
NAME=Cocoon custom

ON=@CREATE
ATTR=ATTR_MOVE_NEVER
//dispid=010da  -------WE MUST ADD THIS LINE FOR FIX IT
```
.add i_cocoon_mephitis1 and you will have the dispid of i_web_cocooon

on 56d it was not like that.

Jhobean:
https://github.com/Sphereserver/Source-X/issues/419

Jhobean:
I think the bug occur in this line: https://github.com/Sphereserver/Source-X/blob/d931bc37042203ec2015107783fdf75c83732d5d/src/game/items/CItem.cpp#L2066

When getting the dispid, don't take the ID= as the DISPID but the "master" ID of this item.

There a difference on 56D and X  on the function returning the dispid

```
	ITEMID_TYPE GetDispID() const
	{
		return (ITEMID_TYPE)(m_dwDispIndex);
	}
```
Item are define differently and I think the bug come from there

```
	ITEMID_TYPE GetDispID() const
	{
		return static_cast<ITEMID_TYPE>(m_wTileID);
	}
```

Jhobean:
Don't seem Fixed. 
[...]

Issue #444, opened 2020-04-28
strange behaviour with instances


GitMadKing:
empty world
execute in console
add i_bee_swarm in game
look at info "instances", it has 8.
try goitemid i_bee_swarm
you will go to just the one you created.
even forinstances loop will not loop trough nothing

```
[function f_tsi]
for 0 30
	local.item=<serv.deflist.itemdef.<dlocal._for>>
	serv.log <local.item> <serv.itemdef.<local.item>.defname> <serv.itemdef.<local.item>.instances>
endfor
```

xwerswoodx:
I added a i_bee_swarm. I check it's instances equals 1, later I add another, it says 2, added another it says 3. And used .goitemid i_bee_swarm, first sent me on 3rd one I created, later 2nd and at last 1st one. Seems like everything working fine. Maybe your info script is showing instances info wrong?

GitMadKing:
of course is not info menu.
i use clean sphere, clean xscriptpack, clean saves.
first execute function attached
then do the test you did

GitMadKing:
of course is not info menu.
i use clean sphere, clean xscriptpack, clean saves.
first execute function attached
then do the test you did

even the log in server looks wrong.
every "same item"(dupelists) it increments their value to 8. like there were created.

xwerswoodx:
I tried it and still everything correct for me.

GitMadKing:
@xwerswoodx let me show more detaieled
just downloaded sphere, xscriptapack
THE VIDEO:
https://streamable.com/3j9udg
THE SPHERE:
[clean_sphere_bee.zip](https://github.com/Sphereserver/Source-X/files/4566190/clean_sphere_bee.zip)
```
22:38:5:'Mad King' Says UNICODE 'ITA' 'F_TSI' mode=0
22:38:(book_content_default.scp,10)i_eq_followSrc i_eq_followSrc 0
22:38:(book_content_default.scp,10)i_pagegm_teleport_delay i_pagegm_teleport_delay 0
22:38:(book_content_default.scp,10)i_memory_fishing_splash i_memory_fishing_splash 0
22:38:(book_content_default.scp,10)i_gem_arcane i_gem_arcane 0
22:38:(book_content_default.scp,10)2331 i_bee_swarm 1
22:38:(book_content_default.scp,10)2332 i_bee_swarm 2
22:38:(book_content_default.scp,10)2333 i_bee_swarm 3
[...]

PR #572, opened 2020-12-19
Fixed some issues

xwerswoodx:
**-Fixed: UnlockItem -1 doesn't remove flags from items inside multi.**
**-Fixed: Serv.Itemdef.ID.Instances adding dupeitems as instance. (Issue: #444)**
**-Added: TAGAT to ACCOUNT reference. (Issue: #531)**
**-Fixed: Timerf stop space issue. (Issue: #554)**
  I used Nirad's recommendation, there was some other ways but i thought this was basic solution.
**-Fixed: IDs from dupelist can't able to used as baseid for items. (Issue: #512 & #419)**
**-Fixed: RangeL and RangeH return values. (Issue: #528)**
**-Improvement: Now '@' and SQRT math expressions can able to calculate negative values. (Issue #371)**


xwerswoodx:
I don't know why pull requests keep showing the commit changes, i guess it's because of I merged 3 times while doing my fixes. But you can see changes from first (Fixed unlockitem -1...) and 4th (Some fixes) commits.

src/common/CExpression.cpp
```diff
@@ -249,13 +249,23 @@ bool IsValidGameObjDef( lpctstr pszTest )
 {
 	if (!IsSimpleNumberString(pszTest))
 	{
+		/*
+		 * We need to check if defname is exist first, and get the value to override pszTest lpctstr to let people can add item from defname.
+		 * Check issue #357 for more information (https://github.com/Sphereserver/Source-X/issues/357)
+		 * xwerswoodx
+		 */
+		const CVarDefCont * pDefBase = g_Exp.m_VarDefs.GetKey(pszTest);
```
cbnolok, line 257: I removed that when making RESDEF.
Before that, both DEFNAME section data and item/char defnames were stored in g_Exp.m_VarDefs. That was inefficient, because if you are searching for a DEFNAME, you'll have to search for it between thousands of item/char defnames.
I didn't think about backwards compatibility defs though, so i guess i'll have to add soon a RESDEF section, similar to the DEFNAME one.

src/game/components/CCPropsItemWeapon.cpp
```diff
@@ -71,12 +71,23 @@ bool CCPropsItemWeapon::GetPropertyNumPtr(PropertyIndex_t iPropIndex, PropertyVa
     switch (iPropIndex)
     {
         case PROPIWEAP_RANGEL:
+        {
+        	if (_uiRange == 0) {
```
cbnolok, line 75: Could you please use the code style convention for spaces/tabulations and for brackets?

cbnolok (changes requested):

src/common/CExpression.cpp
```diff
@@ -626,7 +636,17 @@ llong CExpression::GetSingle( lpctstr & pszArgs )
 								iResult = (llong)sqrt( (double)iTosquare );
 							}
 							else
-								DEBUG_ERR(( "Exp_GetVal: Sqrt of negative number (%lld) is impossible\n", iTosquare ));
+							{
+								/*
```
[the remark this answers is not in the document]
xwerswoodx: Hmm, I just get it wrong. I was not a fan of changing SQRT way so i just cancelled changes, i will pull after complete other conversations.

src/common/CExpression.cpp
```diff
@@ -1016,17 +1036,34 @@ llong CExpression::GetValMath( llong llVal, lpctstr & pExpr )
 			++pExpr;
 			{
 				llong iVal = GetVal( pExpr );
+				bool isNegative = false;
```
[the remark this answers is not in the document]
xwerswoodx: For this, the way I did it same as explained here but if power command do it automatically I can change it to default. I will make some tests then pull it :)

src/common/CExpression.cpp
xwerswoodx: Hmm, I see "m_VarResDefs" already exist for Resource Defs, but it doesn't include Defnames inside it, but adding them into m_VarResDefs can cause more problems i think, maybe as you say creating RESDEF is better solution for it. I thought it can be problem to check all vars all the time but "IsValidGameObjDef" boolean only called under .add command so I don't know if it should change but RESDEF could be better for items because item IDs and MORE1 etc also check defnames and every item does thousands of check all the time. But this boolean only called under .add command so this can be temporary solution for .add, if you want I can remove it, but I think this can stay until RESDEF added, then we can change it to RESDEF instead of checking vars. But checking vars for .add, which called by client and not used all the time, won't cause much issues i think.

src/game/components/CCPropsItemWeapon.cpp
xwerswoodx: I haven't got what you mean but I think you mean me to use the rule that move brackets new line, if it's I am already using it, but it's like habit for me to remove, but I can move brackets new line if it would be better :)

src/common/CExpression.cpp
xwerswoodx: I am waiting your answer to change that, i leave that until resdef added, then we can remove it. It's just for ADD command so i think it doesn't cause much issue, but if you want, i can cancel this and pull again :) Other requests done.

src/common/CExpression.cpp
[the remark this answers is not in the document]
cbnolok: Power doesn't do that automatically sadly, because if one of those particular parameters are passed it will instead change the global errno value, you don't know if the operation was successiful just from the return value. It's better to do the checks before calling power

src/common/CExpression.cpp
cbnolok: The problem may be happening only with .add (not sure) because there are other workarounds (checking both DEFs and RESDEFs) in some other sections of the script parsing code, but as said those are workarounds, ideally removed in future, and we shouldn't rely totally on them. I can do some testing and then send the new RESDEFNAME section commit!

src/game/components/CCPropsItemWeapon.cpp
cbnolok: Yup we have bracket style and indentation (use spaces instead of tabs) at the bottom of readme.md!

src/common/CExpression.cpp
xwerswoodx: This is already removed with the latest commit so I think I can resolve this conversation.

src/common/CExpression.cpp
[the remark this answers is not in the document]
xwerswoodx: This is cancelled already.

src/game/components/CCPropsItemWeapon.cpp
xwerswoodx: I done the inline changes and changed all my codes with the new inlines.

src/common/CExpression.cpp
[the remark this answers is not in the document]
xwerswoodx: I checked and see power work fine with negative values, I updated file and pulled again with a new fix, that only cancel negative value error on Sphere.

src/common/resource/CResourceRef.cpp
```diff
@@ -17,7 +17,12 @@ CResourceRef::CResourceRef()
 CResourceRef::CResourceRef(CResourceLink* pLink) : m_pLink(pLink)
 {
     ASSERT(pLink);
-    pLink->AddRefInstance();
+    /*
```
cbnolok, line 20: My doubt here is: what if, instead, one wants to check the number or loop through the instances, including the dupe items? This solution would cause the opposite issue.
An idea is to add a separate counter for strict instances and dupe instances, increased with AddRefInstanceStrict and AddRefInstanceDupe (valid only for items). In this case, FORINSTANCES needs to accept an additional parameter, 0 (default, with dupe item instances)/1 (strict, only direct instances).
I'd like to hear from other devs and contributors here!

cbnolok (changes requested):

src/common/CExpression.cpp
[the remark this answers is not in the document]
cbnolok: Would you like to add the checks on power operands here on this pull request, or do you prefer me to do that on a separate commit?

src/common/resource/CResourceRef.cpp
cbnolok: In any case, reading the issue #444 it looks like we need to decrease the strict ref counter or to avoid its increment when the ID changes

xwerswoodx: I guess that because I haven't explained myself well. FORINSTANCES still working fine, this pLink is called when sphere read dupelist and every item added 1 point in instance, but when I just cancel this, FORINSTANCES and SERV.ITEMDEF.item.INSTANCES still working fine. 

How I tested:
I added i_bee_swarm 2 times.
Flip one of them,
I added one of the i_bee_swarm item from it's dupelist

I used FORINSTANCE loop on i_bee_swarm - returned 3
I used .show serv.itemdef.i_bee_swarm.instances - return 3

I haven't seen any issue on it but maybe I am wrong, I am just trying to get best result, and give best information I can give :)

src/common/CExpression.cpp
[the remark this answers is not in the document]
xwerswoodx: What type of checks? Because I think the only check for error is already there, for negative values seems like pow working fine. Am I missing something?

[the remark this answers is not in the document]
cbnolok: Oh good, i didn't realize that snprintf in CFloatMath::FloatMath handled those special values and i thought we needed manual error checking. Works good then!

Issue #544, opened 2020-11-23
SQL null values cause errors in sphere


orishi:
When you have null values in SQL table and then you try to select those, it causes error in sphere.

Let's make a table:
`CREATE TABLE test_table (some_column VARCHAR(20) NULL);`

Insert some data:
```
INSERT INTO test_table VALUES ('something');
INSERT INTO test_table VALUES (NULL);
INSERT INTO test_table VALUES ('something2');
```

SCP:
```
[function sql_test]
db.query "SELECT * FROM db.test_table"
if (<db.row.numrows>)
	for 0 <eval <db.row.numrows>-1>
		serv.log <db.row.<dlocal._for>.some_column>
	endfor
endif
```

Will result in:
```
16:47:(function_commands.scp,807)something
16:47:(function_commands.scp,807)0
16:47:(function_commands.scp,807)0
```

And huge error list:
```
16:47:DEBUG:Printing STACK TRACE for debugging purposes.
16:47:DEBUG: __ thread (id) name __ |  # | _____________ function _____________ | microseconds passed from previous function start
16:47:DEBUG:(05012)          T_Main |  0 |     CNetworkManager::processAllInput | +0  
16:47:DEBUG:(05012)          T_Main |  1 |          CNetworkInput::processInput | +31292  
16:47:DEBUG:(05012)          T_Main |  2 |           CNetworkInput::processData | +1416  
16:47:DEBUG:(05012)          T_Main |  3 |           CNetworkInput::processData | +3  
16:47:DEBUG:(05012)          T_Main |  4 | CNetworkInput::processGameClientData | +0  
16:47:DEBUG:(05012)          T_Main |  5 |     PacketSpeakReqUNICODE::onReceive | +4  
16:47:DEBUG:(05012)          T_Main |  6 |                  CClient::Event_Talk | +1  
16:47:DEBUG:(05012)          T_Main |  7 |               CClient::Event_Command | +8  
16:47:DEBUG:(05012)          T_Main |  8 |                        CChar::r_Verb | +301  
16:47:DEBUG:(05012)          T_Main |  9 |                     CObjBase::r_Verb | +6  
16:47:DEBUG:(05012)          T_Main | 10 |   CScriptObj::r_Call (FunctionIndex) | +2  
16:47:DEBUG:(05012)          T_Main | 11 |             CScriptObj::OnTriggerRun | +16  
16:47:DEBUG:(05012)          T_Main | 12 |                        CChar::r_Verb | +7  
[...]

PR #574, opened 2020-12-22
Fixed Mysql Null Pointer Issue

xwerswoodx:
Fixed mysql causes debug when the char comes NULL. (Issue: #544)


alexdan2:
Make a little test whit simple table whit null and all work good! No more critical.
Thx

PR #576, opened 2020-12-23
Added ATTR_CANUSE_PARALYZED flag

xwerswoodx:
This flag let people to use items while paralyzed, but still not allow to equip/unequip or moving items, also spell (scroll, wands) or skill (drum, flute) items not allowed to use. This is done to let people to choose some items that can use while paralyzed, like trapped pouch, or some similar type items.

Also while doing this, I added containers inside players' own backpack as exceptional for frozen at all, because in OSI you can do this, so this is default behaviour.

### DIFF
--- a/src/game/items/CItem.h
+++ b/src/game/items/CItem.h
@@ -94,15 +94,15 @@ public:
 #define ATTR_SECURE				0x2000000			// Is Secure
 #define ATTR_REFORGED			0x4000000			// Is Runic Reforged.
 #define ATTR_OPENED				0x8000000			// Is Door Opened.
-#define ATTR_CANUSE_PARALYZED   0x10000000          // Can able to be used while paralyzed.
+#define ATTR_SHARDBOUND			0x10000000
+#define ATTR_ACCOUNTBOUND  		0x20000000
+#define ATTR_CHARACTERBOUND		0x40000000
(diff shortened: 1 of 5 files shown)


src/game/items/CItem.h
```diff
@@ -94,14 +94,16 @@ class CItem : public CObjBase
...
-#define ATTR_SHARDBOUND			0x10000000
-#define ATTR_ACCOUNTBOUND  		0x20000000
-#define ATTR_CHARACTERBOUND		0x40000000
+#define ATTR_CANUSE_PARALYZED   0x10000000          // Can able to be used while paralyzed.
 
 // TODO
 #define ATTR_CANNOTREPAIR		0x400000000000		// No repair, no fortify
 #define ATTR_FACTIONITEM		0x80000000000000	// ? Faction Item (Has cliloc)
 #define ATTR_VVVITEM			0x100000000000000	// ? Vice vs Virtue Item (Has CliLoc)
+#define ATTR_SHARDBOUND			0x200000000000000
```
cbnolok, line 103: Would you mind leaving the old attrs with their value? There's no reason to change them and this could cause problems to people, now or later, if they have used those values for some other stuff

cbnolok (changes requested):

src/game/items/CItem.h
xwerswoodx: Done 👍

PR #591, opened 2021-01-08
Added quote support for strsub to avoid whitespace issues.

xwerswoodx:


xwerswoodx:
I added this support because having spaces from beginning or the end directly trimmed when you use strsub and it causes you remove the characters you don't want.

If I need to give an example
<strsub 1 0  asd > - returns 'sd' instead of 'asd ' (Because I put space at the beginning)
<strsub 0 0  asd > - returns 'asd' instead of ' asd '
<strsub 1 0 " asd "> returns 'asd "' instead of 'asd '

PR #593, opened 2021-01-10
Fixed: Strtoken can't manage to parse arguments which includes more t…

xwerswoodx:
…han one quotes


xwerswoodx:
I tried discard and reset sstring.h twice but it still shown changed at all when I add just new commands inside, I noted here that I have done no change in sstring.h except adding 2 new boolean and int variables.

PR #604, opened 2021-01-12
Fix  the speedhack detection

Jhobean:
Speedhack is a way to move faster. The worse situation is when player set it to 1.1x to 1.2x the normal speed. It's hard to see by other player and because of different ping it' realy hard to see by the server .

On sphere there 2 systems of speedhack detection.
No one of the actual system work and detect speedhack correctly under 1.5x

**First system:** The packet buffer (in function by default)
 This system is base on a big buffer and an average of 7 step (Walk + run). Detection is weird because you don't ignore rotation and you calculate run + walk. It's a sort of big average and not precise. Each time you stop walking the average is reset. So you must be detected in one straight deplacement. In theory you will detect speedhack but under 1.5x it's impossible to detect it. When speedhack is detected your step is rejected but you can custom this events @USEREXWALKLIMIT.

**Second system** The experimental Fastwalk detection
Experimental is the good term. it's just not work on a live server. Each step is evaluate and you check the time it take and it should take. If time is under, your step is rejected withoud the trigger and possibility to let pass the player. On local, it may work if the delay is well ajust ON the core. No ini setting permit to ajust it... At the moment if you have a ping higher than 30, you are undetectable and can speedhack. Admin can ajust the delay on the core but it will permit to detect only low ping. Since it's impossible to have all player the same ping, this system just don't work until you are able to have the real ping on the formula.

**What in this commit:**
1: Rework of experimental system
 - I commented and improve a little bit the second system (Do not work but work better lol)
 - Include speedmode 1 on detection
 - GM are not checked
 - Add a console warning when speedhack detected

2: The big task was to rework the buffer concept and integrate the experimental one on it. We have now the average of the buffer and the precision of each step of the experimental one.

Big modif I done:
- GM are not checked
- Add a dir int on the char for comparing last direction and actual direction . You dont check change direction anymore
- Check only when running
- Check each step instead of 7 (if there is 1 walk step on the calcul, all go wrong)
- Add a console warning when speedhack detected
- Move the fastwalk check AFTER walk verification (step on restricted area or step on player) Because a rejected step was crashing the detection. 

PRO:
- Easy detection of 1.1x whatever the ping
- Keep the same trigger @USEREXWALKLIMIT than before

CONS:
- Ini setting is dramastically change, if someone take new core and do not change ini, all player will stop moving lol
- Since it's more precise, you need some fine tunning of walkregen setting. This setting depend of the ping of server.
If to low, you'll get false positive. If to high, you will not detect anyone.


Jhobean:
Made some improvement and old setting will not cause false positive. So if people change the core and not the ini, it will be transparent to them

cbnolok:
Can you also add a changelog entry? This is an important new feature

Issue #406, opened 2020-04-08
Impossible Kill someone with custom dommage


Jhobean:
We have a little bug with our custom dommage script. Some custom effect apply a DOT ont the char and they lost hit point over time. The problem is if you have 1 hitpoint, and loss 2, your life go to -1 and you do not dye....  After 1 sec, you are a 100% of life.

Here how reproduce it:
```
[FUNCTION test]
hitpoints = <hitpoints> +- <ARGV[0]> 
SAYua 0027 0 0 eng <eval <ARGV[0]>>
```
Juste set the QTE of hit point you need to pass in negative.

Note1: This system was working well on 56d
Note2 : Same on NPC

Note3: If you are GM, dont forget to set .invul to off :-)

xwerswoodx:
Seems like when your hp drops down to negative values, it directly maxed the value, so here is the solution.

```
[FUNCTION test]
LOCAL.HP = <dARGV[0]>
IF (<dLOCAL.HP> > <HITS>)
  LOCAL.HP = <HITS>
ENDIF
HITS -= <dLOCAL.HP>
SAYUA 0027,0,0,eng,<dLOCAL.HP>
```

xwerswoodx:
I checked the code; there is a 2 different version of short values which is **short** and **ushort**. both of 16bits but the difference is short can be set from -32768 to 32767. But on the other hand ushort totally uses positive numbers and can be set from 0 to 65535. This is make the difference. If I change ushort, that lets the hits going to negative and we can check negative values to kill character. But on the other hand, does it really needed? Because normally hits never be less than 0, also the developer who did this change may have his/her own reason or maybe he think same as me, we totally don't need any negative values for stats. So I don't think we need any changes about it, also it lets the others to make npcs with 65535 hitpoints. So It's not totally bug, I suppose you to use that to make sure you have evade that issue;

```
[FUNCTION hpDown]
LOCAL.HP = <EVAL <HITS> - <ARGS>>
HITS = <MAX <dLOCAL.HP>, 0>
```

Or you can use the way I gave you before.

cbnolok:
We need to uniform every signed/unsigned differences in favor of the
unsigned variables, since hits are supposed to be unsigned, this changes
also its maximum possible value

Il ven 24 apr 2020, 11:08 xwerswoodx <notifications@github.com> ha scritto:

xwerswoodx:
[...]

PR #625, opened 2021-01-25
Fixed: Hits,Stam,mana can go in negative (Issue #406). Stacked Deed only consume 1

Jhobean:
Since the Stat_SetVal was an unsigned, negative value was impossible so -10 = 65550. It was impossible to make a check if value was negative.

Because the ival is now signed we can check if it's negative and ajust it to 0.

Because Ushort was able to go to (0, 65000). If we don't want change the functionnality, we must now use int (-65000, 65000)

Fix for issue #406

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -2434,4 +2434,4 @@ Must have event, even if it's empty, since it's applied by the source to generat
 -Fixed #600: Changed old dating method on bboards to now have actual dates. (This affects dates of old messages as the timestamps is different)
 
 24-01-2021, Jhobean
--Fixed: Impossible to kill a char by setting his hitpoints in negative. ex: If hits was 10 and a script do hits -= 10. (Issue #406) 
+-Fixed: when hits, stam or mana was set to negative value, overflow occur and stat was reset at max. Now a negative value will set stat to 0. (Issue #406) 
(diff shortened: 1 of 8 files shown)


Jhobean:
I know @xwerswoodx was saying it's better to keep ushort because we don't go in negative. In an other hand if you set a negative value on hits, it's evident you should die.  By doing the restriction in the server, some old script will be compatible.  

For example on our shard we added custom damage by doing src.hits -= 10 . All the fix are done in my script but we have the problem to on some custom item droping the hits. So player was using the item to drop below 0 and heal them full heal...

cbnolok:
I prefer to keep that ushort for a lot of reasons (one of them: memory usage), also you should ideally do the check in scripts. If we really want to set negative values to kill chars, and we should talk about this, we can just do a check when setting the HITS keyword

Jhobean:
"When setting the hitpoints" its in core or in script?

Jhobean:
I try to track where the stat are calculated before the function set_val for making the check there. This calcul seem to be in parsing function and did not find it.

cbnolok:
Search _HITS:, you can do a pre-check before assigning the value. If the int value is < 0, set that to 0. You can also use std::min for that

Jhobean:
I changed the method. What do you think?
If number is negative, I force to 0

Jhobean:
I added a little fix about t_deed. On my custom shard, t_deed can be stack. But when we use it, all the stack is consume.

PR #717, opened 2021-06-22
HOUSEDESIGNCOMMITITEM 

alexrmrm:
@HOUSEDESIGNCOMMITITEM add:
ARGO	The custom multi will add item.


fjgo86:
The 'fix gold' commit fixes the gold command not allowing you to set the value to 0.
The other adds the house as ARGO in the @HouseDesignCommitItem trigger.

PR #736, opened 2021-07-17
#HOUSECUSTOM:

alexrmrm:
-ADD continueCustomize


Jhobean:
What this feature do?

alexrmrm:
that will resumen last design if player close dialog.

now if you close dialog and make again customize, will start again with last commit, with that can continue with last design maked without commit.

PR #746, opened 2021-08-01
DIALOG INPUT

criminalx:
Implemented a new dialog section "input" - Provides input processing before button triggers. This section can be utilized to check if the input is valid (obj in range, line of sight, etc..). This should reduce amount of code in each button trigger.


criminalx:
Added example in changelog

Jhobean:
Nice feature!

cbnolok:
I'd change the INPUT keyword to something more "meaningful" like PREBUTTON or something like that. Any thoughts?

criminalx:
Perfect boss

Issue #835, opened 2022-02-07
Container max weight not checked when dropping item on a stack


Jhobean:
How to reproduce 
1- set BankMaxWeight=2000 on the ini
2- Open your bank
3- .add i_ingot_iron 10000
4- slide 10000 ingot on your bank (You will be correct because you are a GM)
5- Switch GM OFF
6- Try to add 1000 ingot on your bank by sliding it on the container. (this action should be refuse because bank is full)
7- Try to ad ingot by adding the ingot in the stack of ingot already in your bank. it will work

This check shoud be made: bool CItemContainer::CanContainerHold  but I dont know where to add it

Jhobean:
OK I found the reason. It appear on this commit (https://github.com/Sphereserver/Source-X/commit/0268aad46a4487ea3cf91759bdb036e65b3e0080#diff-9470427737b1c702b50280e674ada7dc4cbbb2fb948a91922ea030ef3030676bL377-L382) because we remove these line on that file: src/game/clients/CClientEvent.cpp
```
else if (!pChar->CanCarry(pItem))
			{
				SysMessage(g_Cfg.GetDefaultMsg(DEFMSG_MSG_HEAVY));
				Event_Item_Drop_Fail(pItem);
				return;
			}
```

When I add it back, it repair the problem but I can't anymore go over my maxweight

Jhobean:
Fix on Xun's Commit: https://github.com/Sphereserver/Source-X/commit/48d50128727253bb90d6be6505b976046453170d
Thx

Issue #822, opened 2022-01-10
Weight: Paperdoll, Dragging, and Max Carry Issues


Soulless-1:
This is a multi faceted issue, but it all ties into one big problem that's constantly reported on many shards over many years.  (prepare yourself for a long paragrah)

Many problems that are due to several issues :  Dropping items when they unequip from your paperdoll and youre overweight,  Not being able to put things in your bag over your weight limit, Not being able to drag heavy items, and not losing stamina properly when overweight

The way UO Does it, and this is why these issues don't exist on the official servers:

Your paperdoll does not count against your weight that can go in your bag, equipped items add weight to you, but you can always equip an item regardless of your weight, should you cast a spell and it cant fit in your backpack it would just bounce back to your paperdoll and you would not drop it on the ground. you should never lose an item because it was unequipped by a game mechanic. and weight in your backpack is the only thing that should determine if you can put something into the backpack. this is the message you get for that :
![image](https://user-images.githubusercontent.com/17236327/148723649-511b4e20-77b9-47c4-b4ad-24f59d36cdd2.png)

Now the backpack's weight limit is based on your strength, and that's the amount of weight until the backpack is overloaded. when the backpack is overloaded it does not mean new items cannot go into it, that means you will lose stamina at a rate of 1 stamina for every 5 stones you are over the max carry weight. this amount you are allowed to have over the weight is about 30 stones over the backpacks max carry weight. at 100 str your backpacks max weight is 390. You can overload it up to 420 (not counting a characters equipped items) When you place an item in the backpack and it has become over loaded it's supposed to tell you that it's overloaded, once you reach 30 stones over the max carry limit then you are no longer able to fit items into the container and they bounce back to wherever they came from. overloaded message :
![image](https://user-images.githubusercontent.com/17236327/148723672-74423d00-8c07-4eda-9d9a-edb54c8536ac.png)
![image](https://user-images.githubusercontent.com/17236327/148724016-0812fbdb-44ba-452d-868e-5636a2cfc368.png)

You should be able to drag any amount of weight/items with your cursor, the weight is already added to your char as soon as you pick it up, so the same mechanics above that dictate weight/movement. If the weight is beyond what you can fit in your backpack beyond the overload limit, then you are not able to put it into your backpack. if is under 30 stones over your max backpack carry limit then you are able to fit it into your bag. Limiting what people can drag only serves as hindrance to organizing banks and chests with heavy items in it. no picture needed for this.

[...]

PR #836, opened 2022-02-08
Updating Weight behavior to be able have old feature and OSI Feature

Jhobean:
With These commits, it will be transparent for actual shard and no transition is require for them. If admin want to improve their weight system, they will be able to do it with the ini settings.

Recent reworked/imported commit from 56d created 2 small bug need to be fix. (Can be done after this commit merging because bug are alteady present)

1- https://github.com/Sphereserver/Source-X/issues/835
I don't really know how to fix it and keeping the BackpackOverload setting

2- The file receive.CPP was rework and since this rework I was not able equip an item by dragging it on the char. I reverted the commit and only commented the restriction of "canCarry" (useless now)
https://github.com/Jhobean/Source-X/commit/0268aad46a4487ea3cf91759bdb036e65b3e0080#diff-cf63904383ebd2f4524a8cbad029370aa5eee5f168f0c6b924770432a31b9368L542-L552


Jhobean:
Should fix this issue
https://github.com/Sphereserver/Source-X/issues/822

PR #880, opened 2022-08-01
Bunch of updates.

JulianUO:
Changelog for the changes included in the pull request:

- Added: t_talisman for talismans. These items should be equipped on layer_talisman. Remember to set the layer and add some cool functionality in the scripts :)
- Added: t_armor_chain and t_armor_ring typedefs. You can still use t_armor for every piece of armor, but also allows you to use these new types to use specific behavior for each type of armor.
		 *IMPORTANT* If you want to use these types, you need to add the following definitions under [TYPEDEFS] in your scripts:
			t_armor_chain			302
			t_armor_ring			303
			t_talisman				304
- Added: SPELLFLAG_FREEZEONCAST, when a player casts this spell it gets frozen, opposite to SPELLFLAG_NOFREEZEONCAST 
		 Example use: freeze while opening a gate or recall). MAGICF_FREEZEONCAST must be disabled.
		 *IMPORTANTE* You need to add the following definition under [DEFNAME spell_flags] in your scripts:
			spellflag_freezeoncast	   0100000000 
- Added: Chars TAG, TAG.SkillMod%d being %d the SKILL ID. This allows you to increase the bonus for certaing skill instead of modifying the raw base value (just like stats affect skill bonus).
		 Example: TAG.SkillMod10 = 100 will bonus you 10.0% in Camping (SKILL 10). If you click on "show real" in the skill menu, you won't see this bonus but the real value.
		 Note that most of the internal calculations are used taking in account stats + tag bonus.
- Added: @SpellInterrupt, fired when a char interrupts someone's casting. Default is the interrupter, SRC is the person casting.
		  ARGN1 = Spell being casted.
		  ARGN2 = The disturb chance calculated: 0-1000. (Internally, this number is compared against a random of 1000)
		  RETURN 1 cancels the interruption and allows the target (SRC) to cast.
- Added: SKILLUPDATE %d, argument is the skill to update. Use this function with the tag described previously.
- Added: [sphere.ini] ColorNotoGoodNPC, you can now set a custom color for good npcs (this should not affect tamed creatures).
- Changed: ATTR_QUESTITEM to have internal functionality. These items can't be dropped, traded or looted (as described in script definition).
- Changed: ATTR_CANNOTREPAIR to have internal functionality. These items can not be repaired.
- Changed: Mining @Start/@SkillStart, now ACT is the Worldgem Bit we're mining from. You can modify it easily instead of using workarounds to get the UID.
- Changed: @UserQuestArrowClick, using return 1 now prevents from showing default message when (left/right)clicking arrow.
- Changed: Crafting skills now won't display the bounce message when successfully crafting an item, with the skill message is enough.


cbnolok:
Which typedefs would you use for armors in the default scriptpack? Switch the chain and ring to the new ones? Though, a lot of typedefs for every armor are missing... In any case, i wouldn't touch the armor typedefs on the source, if someone needs more customization he (or we, in the scriptpack) could add a custom softcoded typedef as a tevents on the armor scripts

JulianUO:
Aye, what I did here is just adding the basic armor types which (for me) are plate, ring, chain and leather. You can still use the t_armor for everything, there's no need to change the scriptpack at all, but if you prefer you can split them into the different types. Never thought about the tevents solution. 

Anyway is just a few lines change (adding the definition and treating the types as armor), I can keep that commit out.

Jhobean:
What I understand, Adding these type on the server side change nothing because action is always the same for T_armor or T_armor_chain. But, it's more evident like this that these kind of armor are different. I think we could implement it like that and eventually if someone want improve something the base will be there.

Heavy armor and medium armor was always the same and should be different. If it's merge, I could modify Script pack to reflect this and be more usefull for admin.

Yes it could be done via script but server can make the first and major division for important categorie.

cbnolok:
I'd like to hear from others for this, my main perplexity is about the need of adding additional checks for the other typedefs in scripts. Though, i have to admit that typedefs for weapons are already fragmented, so it wouldn't be so much of a difference in the big picture

drk84:
For me it's ok, even if i don't think the additional typedef t_armor_ring and chain are necessary.

Jhobean:
> For me it's ok, even if i don't think the additional typedef t_armor_ring and chain are necessary.

You are right, the type is not complete so it's weird to implement like this (but for me it's better than before). What t_armor we will use for the leaf armor? If we add the type, we should take more time and corretly define them all? Add T_armor_heavy?  And keep t_armor for all other?

drk84:
I think the original idea for the two armor typedef was to differentiate between metallic and non-metallic armour

drk84:
Why not add the mining worldgem part also to lumberjacking, fishing etc? or better to all skill with SKF_GATHER

PR #881, opened 2022-08-03
Updated MySQL Connector to 8.0.30

JulianUO:
I leave it here as a Draft and WIP. So everyone can check/comment/review it.

Basically I took the source code from MySQL Server 8.0.30.0
There's a reference to C++ connector on the documentation but the code is unrelated to the one Sphere uses. The main problem is that 8.0 isn't released for 32 bits, only 64 bits. But inspecting some stack overflow some people reported that the .dll from 5.7 works just fine.

The dlls have been updated to 8.0 and added some other needed dlls (libcrypto and libssl).
May need to update README and Documentation.

Rasmus, on discord, reported it works fine but haven't been able to test for myself (lack of time and some other things, specially to check the backwards compatibility.)


cbnolok:
Was it tested then?

Jhobean:
Yes by 2 differents admin. These admin was complain the old sphere did not work with 8.0.  they used this version of .exe and all work well.

On my side, I tested it on my "old" sql and work fine

cbnolok:
Could you add some details in the changelog about where you took the code, which files were added and why and a compatibility list with mysql versions? Is it compatible with mariadb? The first one is very important, because as i understand you aren't using the c connector code, which afair we were using

cbnolok:
I might be wrong, but it looks that you included maybe a big part of MySQL code, i think only some of this is actually needed for our purposes (this could be true also for the new dlls added). Can you investigate further?

PR #884, opened 2022-08-15
Fix listing components by duplicate and houses lockdowns

JulianUO:
By mistake I forgot to remove boat components from being listed twice on the item loops. This fixes it.


JulianUO:
Houses were also having a mix up when locking/securing containers, and releasing them crashed Sphere or generated an exception, since it tried to remove from the wrong list of items or removing an entry that was twice on the same list.

Issue #467, opened 2020-05-10
BRAIN BERSEK + sleep mode cause server fail


Jhobean:
EDIT:
Channge the name of issue since i'm sur it's caused by bersek brain.

I have a recurrent bug on my console:
ITEM go UP continu infinitely
```
10:05:WARNING:1225 items at 2622,2178,2. Sector too complex!
10:05:WARNING:1226 items at 2623,2177,2. Sector too complex!
10:05:WARNING:1227 items at 2622,2177,2. Sector too complex!
```
I found the half of the bug.

At this position, I have NPC with brain bersek hitting an other NPC.  When I port on the position, I see this image with all of blood:

![image](https://user-images.githubusercontent.com/51728381/81501377-0c7d8300-92a6-11ea-944c-edc3edcb7cfb.png)

Instantely all blood dissapear and no more error on console.  Look like the blood appear and never decay. I'm not able to confirm it, but it seem appear when sector sleep.

I'm not able reproduce on vanilla.

Theorie: All sleep on the sector except the bersek NPC? Probably because he came from an awake sector.

GitMadKing:
are you using latest release? this may have been resolved in c97a6bf0f912acd9ad78eabff2e974ecfbad9fa6

Jhobean:
Yes. And its very intermitent.. it do not happen every time.. i dont know how to test anymore.

Jhobean:
08:55:WARNING:24808 items at 1343,1844. Sector too complex!

Lol 24808 Blood item

Jhobean:
I may have a theory... Every time this happen, there is a brain berserk involved. I notice that brsin berserk do not respect the tile distance to their home. He go freely where he want. Maybe, the npc walk to a sleeping sector and hit sleeping npc...  They can defend themselg and never die....  When I teleport, the npc just die in my face and sometime the attacker are already leaved.

Jhobean:
I have an other situation where this bug occur. I habe an npc with a memory. The memory is a timer of 5 sec. Each time the timer go to 0 , blades appear 1 sec near npc and dissspear...   Whem the sector sleep, these item apear but never dissapear....  So same problem. Sector too complex until a player wake this sector.

Jhobean:
The problem was
1- Sector do not sleep if a player disconnect on it (already fix)
[...]

PR #892, opened 2022-09-09
Add @SendPaperdoll, add Decay ITEM never sleep

Jhobean:


cbnolok:
Could you add more details in the changelog about when the paperdoll is requested?

Jhobean:
I reworked the sector sleep function to avoid ALL attr_decay ITEM to never sleep. This will avoid to have useless item stuck at TIMER=0 on sleeping sector.

This change will fix this issue too : #467

Jhobean:
@cbnolok Would you like to take some time to check this and/or comment plz.

cbnolok:
I'd use the cansleep method to check for the decay flag

Jhobean:
OK @cbnolok I made it. i was looking cansleep method but it exist only for sector. For item it's cantick. So I just considere attr_decay on the same way of item with flag CAN_No_Sleep.

PR #894, opened 2022-09-11
Little Fixes (Non Tested yet)

Tolokio:
Non tested yet so this isnt a real Pull Request, I`ll start testing when I have a bunch of fixes, but if anyone is interested in accelerate the proccess, just compile my fork and test it. Dont forget feedback.
Here I will update fixes updated in my fork and can discuss about the fixes.

FIXES:
-Allow vendors use layer_special and keep bought items until next restock. 
-Prevent horses from being killed while ridden. (z=127)
-Fix and unchain archery butter to be used with any weapon with flag SKF_RANGED at 16 tiles max. (only works at 5-6 or if u are close enought to grab ammo.
-Command "Sleep" will wake up chars who already are sleeping.
-Rideable death horses fix.

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -3035,3 +3035,7 @@ Note: The only way the server has to know if the bankself is closed is to store 
 - Fixed: Removing a secured container from a multi didn't work. (Issue #685)
 - Fixed: Moving multis now will add components as first priority to avoid dropping them. (Issue #821)
 - Changed: Skills with SKF_GATHER, @Start/@SkillStart, now ACT is the Worldgem Bit we're gathering from.
+
+
+11/09/2022 Tolokio
+-Removed check that denied vendor to use layer_special to keep bought items. Only pets vendors were allow. (should be the opposite maybe?)
--- a/src/game/chars/CCharAct.cpp
+++ b/src/game/chars/CCharAct.cpp
@@ -3052,7 +3052,7 @@ void CChar::SleepStart( bool fFrontFall )
 	if (IsStatFlag(STATF_DEAD|STATF_POLYMORPH))
 		return;
 	if (IsStatFlag(STATF_SLEEPING))
-		wake()
+		wake();
 		return;
 
 	CItemCorpse *pCorpse = MakeCorpse(fFrontFall);
--- a/src/game/chars/CCharAct.cpp
+++ b/src/game/chars/CCharAct.cpp
@@ -2766,6 +2766,8 @@ bool CChar::Horse_Mount(CChar *pHorse)
 	LayerAdd(pMountItem, LAYER_HORSE);	// equip the horse item
 	pHorse->StatFlag_Set(STATF_RIDDEN);
 	pHorse->Skill_Start(NPCACT_RIDDEN);
+	pHorse->SetTopZ(-127);
+	//pHorse->SetDisconnected();
 	return true;
 }
 
--- a/src/game/chars/CCharAct.cpp
+++ b/src/game/chars/CCharAct.cpp
@@ -3093,6 +3093,17 @@ bool CChar::Death()
 			return true;
 	}
 
+	if (IsStatFlag(STATF_RIDDEN))
+	{
+		CItem* pMountItem = m_atRidden.m_uidFigurine.ItemFind();
+		if (pMountItem)
+		{
+			CChar* const pRider = Horse_GetMountChar();
+			if (pRider)
+				pRider->Horse_UnMount();
+		}
+	}
+
 	// Look through memories of who I was fighting (make sure they knew they where fighting me)
 	for (CSObjContRec* pObjRec : GetIterationSafeContReverse())
 	{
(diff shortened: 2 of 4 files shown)


fjgo86:
Can you change please the max range for archery butter to avoid using "magic numbers" please? You can use some defined macros like UO_MAP_VIEW_SIGHT or UO_MAP_VIEW_SIZE_DEFAULT https://github.com/Sphereserver/Source-X/blob/master/src/game/uo_files/uofiles_macros.h#L19, VisualRange https://github.com/Sphereserver/Source-X/blob/master/src/game/chars/CChar.cpp#L847 or create a new macro/property if it's really needed.

About mounts being killed, how do they get killed? Mounts may die under certain circumstances (eg: poisoned).

Tolokio:
I dont know what do u mean with "magic numbers". I dont know how to set macros right now or config options at sphere.ini. Tried but I messed all.

it seems that mounts are affected by area and walls damage. So they can be killed where they stay waiting. 
Mounts from stables gets disconnected so dosnt get damage I think.

Tolokio:
image

return true better?  butte will not work until u are at 5 or 6 tiles from for shooting and <=2 tiles for grabbing ammo.

fjgo86:
My point regarding "Magic Numbers" is to not have any numbers that doesn't have a name, meaning that "16", you can use instead : return iDist <= UO_MAP_VIEW_SIGHT; <--- this UO_MAP_VIEW_SIGHT is a macro, if you follow it you will se that it's used to set the maximum view range for creatures, which makes more sense to be used since it would make the range for this item equal to the maximum range for almost every other things and prevent having distances above the scope or being to dispairing.

Also, a clan flag was added some time algo to prevent distance checks, so you can add your custom distance in the @DClick: can_i_dcignoredist, isn't it working?

Tolokio:
Can use uo_map_view_sight if u prefer. 

The change is for two reasons, unblock  throwing weapons on butte, and there is an issue related to this. I think cos typedef is not taken right from duped items.
Can be fixed by Can_i_dcignoredist I think, but this is not about that.

The distance check on canITouch() is not neccesary, cos it is duplicated. Butte behaviour will check distances once again at his turn, and will check as well if weapon is ranged, not only distance. Cheking two times for the same thing is not good done I think. Even there was no problem, range and distance checks should be removed so sphere dont waste resources doing two times same thing.

The best solution would be to remove the entire case and include can_i_cignoresdist in the x-scriptpack. But servers that are already running would need to change scripts to let buttes work.

drk84:
The horse fix is not acceptable!!

Tolokio:
> The horse fix is not acceptable!!

Drk, Dont say it is not acceptable without explanation plz. I want to learn. 
Why it is not acceptable? Area effect will still affecting mounts even underground? I did not test it yet but could be.
Others fixes are acceptable then? xD

I dont find others options better than this. And it just add a very light line of code... IF works I think it is weird but clever.

Other options may require much more code. Doing the horse invul, stone or disconnected is not fine for me and many others. Mounts should be able to receive damage passing @gethit, and custom timers should work while ridden.
 What's left?

Jhobean:
Xun suggested something by script on the horse issue. Maybe a combinaison server + script.  On my side I think like dark -127 is not acceptable because it's a not elegent workaround. And sometime it should not work.

Tolokio:
I think it is not about elegant it is about performance.

The unmount can be done by @death ?

I will think other light solutions for the damage. But I think z-127 it same acceptable as leaving the mount at the ground....
Nobody gets closes to that z on their maps.

cbnolok:
As a general best practice guideline, it's better to send a fix per single pull request, otherwise discussion of a fix blocks also the others.
How would the z trick work? Only getting the mount out of touch? In any case it's a workaround and not the best solution, which may be to block an attack on the HitTry stage if the npc is in a mounted state.
For the archery butte, that fix might work for your shard but isn't a good generalization. Apart from using a constant macro instead of a plain number, i think that the best solution would be to check against the npc range, there should be a method already coded afair

Tolokio:
Hi nolok, u are right, one fix one pull. Will be done right next time. But This is not real pullrequest. I just opened this to talk and learn. I recently started to understand source and I want to help with issues. 

About horses: Im looking for that hittry stage, but I dont see how it deals with area damage or fire fields. t_fire for example pass damage directly to gethit, I dont find any stage process from t_fire to being damaged.


My sleep fix is for "src.sleep"  or ingame command .sleep for example. Not for sectorsleeping. but I dont find CChar::CanSleep. Only gosleep(), not sure where to add the check or why.

Would this be fine for the archery butte? 
`
case IT_ARCHERY_BUTTE:
        {           
            const CItem * pWeapon = m_uidWeapon.ItemFind();
            if (pWeapon) 
            {
                    return (iDist <= pWeapon->GetRangeH());
                    //or
                    return (iDist <= GetRangeH() //this better??
            }
        }
`

I think someone tested that char dosnt get unmounted if horse dies.

Tolokio:
Tried to create a separate pull request but I dont know how and now it is done :/

Added a fix for rideable death horses. I dont know if is fine for sphere standards. Not tested, just waiting feedback about the code.

Jhobean:
For different pull request, you must work on different branch.
About the "sleep fix" you want to add a toggle? If you try sleep someone already sleeping it awake him.  What about stone? How doe it work

Tolokio:
About the sleep fix, sleep already uses args for other thing, but Im trying to make wake() avaible for be used by script. 
image
So instead .sleep 1 or 0 u could use .sleep or .wake, I think this can replace the first fix.

PR #900, opened 2022-09-19
Fix: Vendors dont keep bought items

Tolokio:
Only player's vendors were allow to keep bought items. As I dont see any issue about player's vendors using it, I just removed the check.

### DIFF
--- a/src/game/clients/CClientEvent.cpp
+++ b/src/game/clients/CClientEvent.cpp
@@ -1539,7 +1539,7 @@ void CClient::Event_VendorSell(CChar* pVendor, const VendorItem* items, uint uiI
 		}
 		else
 		{
-			if ( pVendor->IsStatFlag(STATF_PET) && pContExtra )
+			if ( pContExtra )
 			{
 				CItem * pItemNew = CItem::CreateDupeItem(pItem);
 				pItemNew->SetAmount(amount);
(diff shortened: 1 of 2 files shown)


src/game/clients/CClientEvent.cpp
```diff
@@ -1539,7 +1539,7 @@ void CClient::Event_VendorSell(CChar* pVendor, const VendorItem* items, uint uiI
 		}
 		else
 		{
-			if ( pVendor->IsStatFlag(STATF_PET) && pContExtra )
```
cbnolok, line 1542: Instead of changing this long-present default behavior, could you add a ini flag to enable it?

cbnolok:
Is that happening only for player vendors or also for npcs?

src/game/clients/CClientEvent.cpp
Tolokio: > Instead of changing this long-present default behavior, could you add a ini flag to enable it?

Never did but I could try. Anyway I would like to know why player's vendors are allow to use the pContExtra. It is weird


> Is that happening only for player vendors or also for npcs?

Players vendors are allowed but I think they shouldnt.
 I think someone introduce this "long feature" while trying to deny player's vendors to use pContExtra. But he did the opposite and forget. OR does players vendors need pcontextra?

Tolokio:
> 

npcs

PR #903, opened 2022-09-22
Fix: Rider dont get dismount if mount dies.

Tolokio:
add: unmount rider if mount dies


Tolokio:
No fix needed. it should happen at CChar::NPC_PetClearOwners()

closing

PR #905, opened 2022-09-23
Fix: Mounts losin action 111 if damaged.

Tolokio:
Damaging a ridden mount will set its action -1. This causes mounts to stop ticking.

Fix would just avoid: fighting back and dessert if were harmed by master.

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -3038,3 +3038,9 @@ Note: The only way the server has to know if the bankself is closed is to store 
 
 19-09-2022, Tolokio
 -Added: Wake() is now avaible to be used by script using command: "wake"
+
+
+23-09-2022, Tolokio
+-Fix: Mounts losing action 111 when harmed by anyone.
+Now they dont try to fight back or desert if is hardmed by own master while being ridden.
+Mounts where unable to die or tick if action 111 were changed. Now seems ok.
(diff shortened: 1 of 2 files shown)


cbnolok:
Can you bring an example of a ridden mount being damaged by the master? To being dismounted or not is something some shards might want to customize. Also, have you tested this extensively?

Tolokio:
> Can you bring an example of a ridden mount being damaged by the master? To being dismounted or not is something some shards might want to customize. Also, have you tested this extensively?

Example: Someone or the master poison a/his horse, then someone or you mount the horse. At first tick horse will be damaged. Statf_ridden will be lost, and horse stop ticking. So poison get stopped even if u dismount. And as horse cant tick, the horse become inmortal and only will pass CChar::death() (@death) when dismounted. So u can´t fix it by script.

This fix dosnt dismount the char, there is a code that should dismount it, but does not work. When code reach that part, mount have lost staft_ridden so no dismount it done(even with this fix). I did another pull about that. This fix just make ridden mounts able to tick if gets damaged. Nothing more.

Not really deep tested but tested and still testing.

Statf_ridden (action) is removed when pet_dessert or pet_release (before dismounting it), it is also removed at clear_all() and other function related to attackers, it also gets removed if try to fight back. This fix avoid this to happen. But if pet dessert by food, it will lose statf_ridden and ticks.

This fix need more fixes by script or source. But this would be the first of all.

Tolokio:
I tested this deeply, it works fine for me.
There is another pull request that complement this fix. If only this pull were implemented there will be issues but the issues can be fixed by script using @death trigger just like this:
```
on=@death
if <flags>&statf_ridden
return 1 //deny death while being ridden
endif

//or

on=@death
if <flags>&statf_ridden
f_find_rider_and_dismount //dismount char before mount dies.
endif
```

Take in mind, right now it is not possible to use @death trigger on ridden mounts. Mount will not pass by @death trigger until player manually dismount because if a ridden mount is harmed, ticks will be denied for that mount until char gets dismounted.

PR #908, opened 2022-09-23
Fix Not dismounting if pet_release() or death()

Tolokio:
This fix makes mounts dismount their riders if mount dies or get released some how. (note this fix needs a related fix to properly work: https://github.com/Sphereserver/Source-X/pull/905

Reason for this fix: if mounts dies, Char needs to be dismounted before corpse is created or it will glitch.
 There are others ways to achieve it but I think this is the best method. (At least for me)

known issues:
-Mount dosnt show death animation (falling anim) when it dies, it just appear dead on the floor wihout any anim. 

To avoid getting dismounted if pet desserts or dies while being ridden, just:
```
@death/pet_dessert
if <flags>&statf_ridden
return 1
endif
```

### DIFF
--- a/Changelog-X1-Nightlies.txt
+++ b/Changelog-X1-Nightlies.txt
@@ -3038,3 +3038,7 @@ Note: The only way the server has to know if the bankself is closed is to store 
 
 19-09-2022, Tolokio
 -Added: Wake() is now avaible to be used by script using command: "wake"
+
+23-09-2022, Tolokio
+-Fix: death and dessert of ridden npcs now dismount npcs.
+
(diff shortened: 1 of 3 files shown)


Tolokio:
Tested deeply, no bugs finded by my side but death animation not being played on the mount when it dies., Mount will instanly appear dead on the floor without falling animation. Still working on it.

But this fix need the other pullrequest I did to work properly. Right now if a ridden mount is harmed, it stop ticking until the rider dismount.

The dismount feature can be denied by script using this script:
```

on=@death
if <flags>&statf_ridden
hits=1 //give hp to the mount so this trigger dosnt fires again and again.
return 1 //deny death so rider will not be dismounted.
endif

//or

on=@gethit
if <flags>&statf_ridden
return 1 //deny mount getting harmed.
endif

```

JulianUO:
Why don't you put the if code from CChar::Death() inside NPC_PetRelease() just before SoundChar(CRESND_NOTICE);?
It works fine for me. Animation is played but it is too quick to see.

Tolokio:
npc_petrelease() dosnt get fired when a mount dies, But will be fired if the owner harms the mount, for example using .kill command.
Using ".xhits 0" instead .kill is the right way to test the issue.

Animation issue still happening.

PR #920, opened 2022-10-18
Changed TimerCall description to use seconds instead of minutes

Raylde:


cbnolok:
This solution can be problematic for backwards compatibility... We can add instead a second argument, like precision level: if missing or 1, the first number means minutes, if 2, seconds (and in the future, if ever needed, 3 for milliseconds). We have to worry first to not break other people builds

Raylde:
> This solution can be problematic for backwards compatibility... We can add instead a second argument, like precision level: if missing or 1, the first number means minutes, if 2, seconds (and in the future, if ever needed, 3 for milliseconds). We have to worry first to not break other people builds

is not a big problem, i think, if other shards need to use minutes instead of seconds, only change 1 to 60, the other way we need another variable and checks.
Right now the developers only need to change 1 number

cbnolok:
Yes but not every developer carefully reads all the changelog, and the worst thing to experience is having few more subtle bugs after updating sphere without having errors.

Jhobean:
I'm agree with nolok. I just think of someone have an XP system with this function. The admin change the .exe boot his server and notice the difference 1 week later after someone have 100x more xp than suppose. The rollback will be horrible. Second argument, or something similar, is a good idea.

Raylde:
Now with the change of @Jhobean should be done.

PR #932, opened 2022-10-20
Last added waittime 2nd argument trigger has been edited.

canerksk:
The coding is also written as an argn2 start trigger, but when I examine the codes, it seems that it is added to the prestart trigger, not start.


drk84:
While they use the same **CScriptTriggerArgs pArgs** the 2nd argument is defined after the "prestart  trigger" ends!

PR #934, opened 2022-10-21
Add CodeQL code scan

Raylde:
Add new workflow to scan the code and to have a tracking of errors and problems

Example 
image


cbnolok:
What needs to be done in addition to make it work? Could you update us with a comment when it's ready for pulling?

PR #935, opened 2022-10-21
Improve build workflow to upload changelog

Raylde:
- Change build pipeline for generate changelog with list of all logs and upload to nightly tag.
- Now the nightly binaries are upload as prerelease


Raylde:
Example 
image

Issue #951, opened 2022-10-29
spellbook in hands (layer1 / layer2) not recognized


AtillaG1:
If i equip spellbook in hands (layer 1 or layer2), the spellbook is not recognized and i cannot cast any spells, as the char or core doesnt sees it.
![image](https://user-images.githubusercontent.com/25097484/198850669-62e6c4b4-c549-495c-95f9-f8729df279e0.png)

Jhobean:
See a proposed fix here: https://github.com/Sphereserver/Source-X/pull/955

PR #954, opened 2022-10-30
Fix: Spellbook in layer2 wasnt seen by Sphere (issue #951)

AtillaG1:
Thanks to @Jhobean and  @Tolokio, they made this fix. Apparently it should be possible to fix it in better way, but this would need someone more skillfull.
I tested it, and now the spellbook works in Layer 1, Layer 2 or from backpack.
#951 

PS: the commits are a bit messy, still figuring out how to work with git and reversing changes, then pushing or doing PR, sorry for that.

🤠


cbnolok (commented):
What about something like this?
CItem* pItem = LayerFind(LAYER_HAND1);
pItem = pItem ? pItem : LayerFind(LAYER_HAND2);
if (pItem && pItem->IsTypeSpeelbook) ...

Also, consider always using const qualifier, when appropriate (like const CItem *pItem = ... if you're using const methods).

cbnolok:
Closing, since discussion has moved on #955.

PR #955, opened 2022-10-30
Added possibility to cast with a spellbook on layer 2

Jhobean:


drk84:
The best way is to  copy paste the spellbook1 code,  because with this commit there will be a problem in the very rare case where two spellbooks are wielded in both HAND1 and HAND2.

The spellbook in HAND2 always take precedence and if it does not contains the spell we are trying to cast, the spell will never be casted even if it's present in the spellbook in HAND1.

Jhobean:
@drk84 true!!!  Tolokio had the good way:smile: i'll fix later today.

Tolokio:
image

Tolokio:
CItem* pItem = LayerFind(LAYER_HAND1);
pItem = pItem ? pItem : LayerFind(LAYER_HAND2);
if (pItem && pItem->IsTypeSpeelbook) ...

This wat if u got a weapon on layer1 code will never look for a book in layer2, or am I wrong?

cbnolok:
Exactly. That's why i was suggesting in the other issue (which at this point is better to close, since it's a duplicate) to expand that code to cover that issue

AtillaG1:
> @drk84 true!!! Tolokio had the good way😄 i'll fix later today.

😆

Tolokio:
> The best way is to copy paste the spellbook1 code, because with this commit there will be a problem in the very rare case where two spellbooks are wielded in both HAND1 and HAND2.
> 
> The spellbook in HAND2 always take precedence and if it does not contains the spell we are trying to cast, the spell will never be casted even if it's present in the spellbook in HAND1.

It could be not a rare case, One paladin/ninjtsu/bushido/spellweabing on layer1, and magic spellbook on the other. But I never thought about this. I just pref to do as drk said cos performance.

So what cbnolok suggested would cause the same issue. right?

Good seen drk.

Jhobean:
Better like this?

PR #957, opened 2022-10-31
Houses count fix

Tolokio:
Fix for "remove" and "nuke" not working well with multis.

No script's fix is required now. Tested and working well with houses and ships.

EDIT:

DONE another PR. with better fix. I leave this open until @canerksk answers!!!

### DIFF
--- a/src/game/CObjBase.cpp
+++ b/src/game/CObjBase.cpp
@@ -2395,6 +2395,21 @@ bool CObjBase::r_Verb( CScript & s, CTextConsole * pSrc ) // Execute command fro
 			return pClientSrc->addGumpDialogProps( GetUID() );
 		case OV_REMOVE:	//remove this object now.
 			EXC_SET_BLOCK("REMOVE");
+			if (IsItem())
+			{
+				CItem* pItem = static_cast <CItem*>(this);
+				if (pItem->IsTypeMulti())
+				{
+					CItemMulti* pMulti = static_cast<CItemMulti*>(this);
+					CChar* pOwner = pMulti->GetOwner().CharFind();
+					if (pOwner)
+					{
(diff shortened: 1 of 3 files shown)


canerksk:
Add a coowner to the house and then delete that added coowner character with remove 1, then after removing that house and taking backup, it crashed.

AtillaG1:
could not reproduce it either.

src/game/CObjBase.cpp
```diff
@@ -2395,6 +2395,21 @@ bool CObjBase::r_Verb( CScript & s, CTextConsole * pSrc ) // Execute command fro
 			return pClientSrc->addGumpDialogProps( GetUID() );
 		case OV_REMOVE:	//remove this object now.
 			EXC_SET_BLOCK("REMOVE");
+			if (IsItem())
+			{
+				CItem* pItem = static_cast <CItem*>(this);
+				if (pItem->IsTypeMulti())
```
cbnolok, line 2401: Couldn't this be done in citem::delete?

src/game/clients/CClientTarg.cpp
```diff
@@ -652,6 +652,17 @@ bool CClient::OnTarg_Tile( CObjBase * pObj, const CPointMap & pt )
 
 				if ( m_Targ_Text.IsEmpty())
 				{
+					if (pItem->IsTypeMulti())
+					{
+						CItemMulti* pMulti = static_cast<CItemMulti*>(pItem);
```
cbnolok, line 657: This could be redundant if the check was moved into citem::delete

Tolokio: yes, can be done at to delete(), I prefer it as well.

PR #959, opened 2022-11-01
Add scriptPack on .exe package

Jhobean:
WIth this PR, official X script pack will be added to the .zip file on the FTP. That mean new admin will have only 1 file to download.

YOu can what it give here on my fork: https://ci.appveyor.com/project/Jhobean/source-x/builds/45248453


cbnolok:
Tbh i'd keep them separated, because the scriptpack is only seldomly needed

Jhobean:
When there new admin starting to play, it's always conplicated to show them what download, where to place the file. I tought it could be easier like this.

True 98% of time the pack will be downloaded for nothing and will not be use. It's true for the folder and crypt file too.

Jhobean:
I'll update wiki page to show how set up

PR #977, opened 2022-12-07
Remove missing code

Tolokio:
Tested and working fine.
-Fix; Removed duplicated code
-Fix: @redeed -> return 1 now works as wiki says. (Multi is deleted instead leaving a empty building)
-Fix: Added pDeed check to prevent errors.


drk84:
I think it should be kept in the case return 1 is found in the @redeed trigger, or eventually moved inside:
immagine

with another if (!fAddon)

Tolokio:
image

Wiki says nothing about avoid the house deletion. It just says it avoid the deed creation. How it is supposed to be?

Could we remove the "return;" and let script pass until the end where the muilti is deleted. OR do we fix the wiki? IT seems it was coded just to avoid deed creation, otherwise it would be coded diferent. MAybe someone miss that return; or was a quickfix for something.

What do u think?

drk84:
That return there will stops the method and the house will not be deleted, pratically the multi is not removed, only the addons / components and the ownerships of the multi is cleared. If we remove the DelMulti code you will be still considered the owner of an "empty multi"

Tolokio:
@drk84 
Exactly, but that should be changed to work as wiki says.
Return 1 should only avoid the deed creation, not the house deletion.

Why we want "return 1" to avoid house deletion? HOuse can be avoided by house dialog.

See last commit I did.

PR #992, opened 2022-12-29
Added 2 new ships triggers.

canerksk:
Ship_Move, is triggered when the ship moves, the other, Ship_Stop is triggered when the ship stops.


drk84:
Missin revision file

drk84:
I see two arguments in the trigger: the direction and fStopped. Is  this used when the ships can't move?

canerksk:
It comes from above in the fStopped function. I added it maybe it could be used. But there is no ship stop, ie write events in the trigger. It just sends the direction of the ship to the trigger with read and whether it is fStopped.

PR #994, opened 2022-12-31
Added Multis function for server.

canerksk:
The function is based on the guildstones.* function. It is a function that outputs all items of type t_multi, t_multi_custom and t_ship on the server. It can be used with multis.count as multis.0.name/p/uid/.

Why do we need this command;

If you have an active server and many installed houses/ships on the server, controlling them with foritems creates a lot of processing load. It is also very troublesome to enter the ids of all houses one by one with forinstances. And it is only possible to check with the for loop on the map you are on. I thought it would be easier to control this way, since it is a hassle to control all the houses on all maps, that is, on the server. If you don't think it's needed, don't add it. It has been tested and works flawlessly.

I couldn't separate the previous commit from this commit and it sent both as pull requests at the same time. I'm new on Github. I'd appreciate it if you could help me separate the two commits.


Jhobean:
It's a very interesting feature. Seem to have some compilation error. Maybe because you repush the same mosif of your previous commit.

canerksk:
> It's a very interesting feature. Seem to have some compilation error. Maybe because you repush the same mosif of your previous commit.

If you know of a simpler method that we can control all the houses or ships installed on all maps ie map0, map1, map2, map3, map4, map5, please share.

drk84:
`	{ "MULTIS",					{ ELEM_VOID,	0												}},`

This should be in alphabetical order (after MULFILES)

also

Can you move  this code

`if (IsTrigUsed(TRIGGER_SHIP_MOVE))
    {
        CScriptTriggerArgs Args(dir, fStopped);
        pItemThis->OnTrigger(ITRIG_Ship_Move, &g_Serv, &Args);
    }`

before

`if (fStopped == true)
    {
        CItem * pTiller = pMulti->Multi_GetSign();
        ASSERT(pTiller);
        if (fTurbulent == true)
            pTiller->Speak(g_Cfg.GetDefaultMsg(DEFMSG_TILLER_TURB_WATER), HUE_TEXT_DEF, TALKMODE_SAY, FONT_NORMAL);
        else
            pTiller->Speak(g_Cfg.GetDefaultMsg(DEFMSG_TILLER_STOPPED), HUE_TEXT_DEF, TALKMODE_SAY, FONT_NORMAL);
        return false;
    }`

In this way ARGN2 will be properly output when the ships stops by moving

Issue #956, opened 2022-10-30
Can= Can_o_nosleep do not work like expected


Jhobean:
How to reproduce:

Start a server with sleepsector activate at 1 min and go in any city. Launch the below function "TEST". After awaking the sector (Go 1,1 and leaving) the item should trig each 2 second for eternity.

On correct behavior, Item should continu to tick after the sector is awake and get asleep again.

```
[itemdef i_test]
id=i_gold
type=t_eq_script
Name=Test de sleep
can=can_o_nosleep

on=@create
timer=2
serv.log the UI of the test timer is <uid>

on=@timer
serv.log I ticking!! My sector sleep?<sector.issleeping> cansleep? <sector.cansleep>
timer=2
return 1

[FUNCTION test]
serv.newitem i_test
new.p=1,1
```

Jhobean:
Ok I made some test:

On the commit: [0b1e59c8 ](https://github.com/Jhobean/Source-X/commit/0b1e59c801d23db529080ee46b8b933ee7590054)
CAN_O_NOSLEEP is implemented. The flag seem to work but its a false impression. Sector juste NEVER fall asleep again. The commit implemented a BUG fixed dew week later in this commit: https://github.com/Jhobean/Source-X/commit/50ad5b386305f4a8e71fe4a1f8366a2862130fa4

In conclusion CAN_O_NOSLEEP just dont work or I use it not correctly.

Jhobean:
If we achieve to fix this CAN_O_NOSLEEP. We should implement this feature in parallel of the flag
https://github.com/Jhobean/Source-X/commit/ca673cd6367bb78608bd276c0908ec560c5473bf

LuxionUO:
I can confirm, can_o_nosleep is not working.

Jhobean:
@cbnolok we need your help for this one. I'll pay the beer the night you will work in this 🤣. Just tell me where to ship it.

Jhobean:
Fix here: https://github.com/Sphereserver/Source-X/commit/ffbb62eccfdadc7b75a9625dd1575dfc20262dc4

Thx @cbnolok

PR #1008, opened 2023-01-11
Fix CAN_O_NOSleep and now all ATTR Decay item will never sleep

Jhobean:
Fix Issue #956


AtillaG1:
When will this be commited ? 😍😎

src/game/CSector.cpp
```diff
@@ -191,15 +191,15 @@ void CSector::_GoSleep()
 		CChar* pChar = static_cast<CChar*>(pObjRec);
 		const bool fSleeping = pChar->IsSleeping();
 		ASSERT(pChar->IsDisconnected());
-		if (!fSleeping)
+		if (!fSleeping && !pChar->Can(CAN_O_NOSLEEP))
```
cbnolok, line 194: Can't those checks be moved in the cantick method?

src/game/items/CItem.cpp
```diff
@@ -5964,8 +5964,13 @@ bool CItem::_OnTick()
     EXC_SET_BLOCK("sleep check");
 
     const CSector* pSector = GetTopSector();	// It prints an error if it belongs to an invalid sector.
-    if (pSector && pSector->IsSleeping())
+    if (pSector && pSector->IsSleeping() && !Can(CAN_O_NOSLEEP))
```
cbnolok, line 5967: Same thing here for the checks for nosleep and decay

Jhobean:
I need you help @cbnolok. Can tick method just not work on sleeping sector.

cbnolok:
Solved by the last commit

PR #1017, opened 2023-01-18
Fix chv_kill

Tolokio:
image

I think it is not neccesary to do damage through combat @triggers. It shows emote like: "xxx is attacking gamemaster" sysmessage ->"xxx were killed by GM/server"

There were a bug related to mounts that were hidden when u tested it using ".kill"

The only good thing that probably is the reason it stays like that, is that the damage output is played at player's head.
There is no sendpacket for that?

I think we could remove the damage part and just use sendpackets to show the damage over the char if it is possible. And if not, I prefer to just remove the damage part.

### DIFF
--- a/src/game/chars/CChar.cpp
+++ b/src/game/chars/CChar.cpp
@@ -4291,7 +4291,7 @@ bool CChar::r_Verb( CScript &s, CTextConsole * pSrc ) // Execute command from sc
 		case CHV_KILL:
 			{
 				Effect( EFFECT_LIGHTNING, ITEMID_NOTHING, pCharSrc );
-				OnTakeDamage( 10000, pCharSrc, DAMAGE_GOD );
+				//OnTakeDamage( 10000, pCharSrc, DAMAGE_GOD );
 				Stat_SetVal( STAT_STR, 0 );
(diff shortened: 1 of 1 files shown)


Jhobean:
If you remove the damage the NPC Die? Probably yes because you set hits to 0 line after.

I dont understand what the bug about the mount. Kit was not working?

drk84:
I don't think it's necessary this change

Tolokio:
> If you remove the damage the NPC Die? Probably yes because you set hits to 0 line after.
> 
> I dont understand what the bug about the mount. Kill was not working?

Cos mount was getting inmortal(stop ticks) when someone attack it while being ridden. People use to test that issue killing the invis mount by ".kill" instead ".set hits 0" so nobody realize the problem along many years.
This is not the real reason to change .kill. But it shows that the expected natural behaviour of .kill should be to kill without attacker.
I just think .kill give more problems if it uses an attacker than if it dosnt.

I think this change should be done, @drk84. But I just create this pull to talk about it. 
IT is a minor thing, but u could give any reason when u close it. I would thx it.
Neither do i think it is a neccesary change, but a little improvement.

I propose to remove attacker from .kill.

Whatever is decided is fine for me. But we could fake that this is a democracy and discuss things.

Jhobean:
Have the attacker in log could be usefull if a bad gm kill many creature you have the log

PR #1020, opened 2023-02-06
Optimisation on how spawn point work

Jhobean:
This PR change the behavior of spawn point.
Now default Behavior is spawn point will never sleep until  they finish their job.

I addition to this, 2 new trigger give the possibility to supervise the spawn point.

By default timer -1 will appear when spawn complete. You can bypass this by setting your timer with ARGN on @ADDOBJ


AtillaG1:
you've been busy ! 😁😲

PR #1030, opened 2023-02-12
Visual change on console + SphereCrypt

Jhobean:
I added some visual indication on console base on 56d. I think it's was more clear.

Good feature is warning now admin there no account and show them how to create one.

image


Jhobean:
I added more info about the fix Weirdeness appearing often on boot. I added the name of the item. Maube it could help to find the bug one time.
image

Issue #1029, opened 2023-02-11
f_onchar_delete not called when deleting account


DavideRei:
Just tried with last version windows x64. Clean sphere + clean script pack
![image](https://user-images.githubusercontent.com/118212274/218267035-13d3d0d6-bb88-4a63-bb53-12aee72d3e9f.png)

![image](https://user-images.githubusercontent.com/118212274/218266874-b94ed0f0-b1a2-4408-8cb7-267ed56ec868.png)

![image](https://user-images.githubusercontent.com/118212274/218266887-bdbb96dc-5862-4fd0-ad94-c902d48269d1.png)

![image](https://user-images.githubusercontent.com/118212274/218266935-2ec1ffce-16ce-43e1-98b3-078453daaea5.png)

![image](https://user-images.githubusercontent.com/118212274/218266971-aff6e47b-2a43-485d-9f1e-8a88ccbfc31d.png)

![image](https://user-images.githubusercontent.com/118212274/218267002-967246d9-d7a7-4cef-bb73-eeeea1d6f1b5.png)

Jhobean:
Ok 56d seem to work ok. When you delete an account and there 3 char on it.  f_onchar_delete is trig 3 times.


The difference I see in code id the way the accout is delete
56d use c++ standard method: https://github.com/Sphereserver/Source/blob/b00f063d0f8eec10c56716c2f1e98edffa2013ef/src/graysvr/CAccount.cpp#L50

Use a new array method (STANDARD_CPLUSPLUS_THIS(RemovePtr(pData)))
https://github.com/Sphereserver/Source/blob/b00f063d0f8eec10c56716c2f1e98edffa2013ef/src/common/CArray.h#L516-L519

This new method was impleented by Coruja here:
https://github.com/Sphereserver/Source/commit/a3c101df79aea7a1da9bfddc53b210649e4c37c1

On X it seem to be custom array method:
https://github.com/Sphereserver/Source-X/blob/7af71186748f5059dd11a1dce6be6c3b82272b22/src/game/clients/CAccount.cpp#L223

https://github.com/Sphereserver/Source-X/blob/3dbb44713bec5d720cf3b932970584331637ec18/src/common/sphere_library/CSPtrTypeArray.h#L83-L92

The only reason I see is the c++ method delete all item on the array and delete method for each account is called.


On X, we have a bug when deleting account. Since char are not correclty deleted, when starting the shard after account deletion, all char item have a mislink bug
[...]

PR #1031, opened 2023-02-12
Fix f_onchar_delete not launch when delete account

Jhobean:
Issue #1029 

Need to be review. DeleteChars() was already on this function:
CAccount::~CAccount()
But this function seem to never be use so I added the DeleteChars(); in the account delete function.

Since we want a notification in script, I mannualy add NotifyDelete(1) and add a bool value to block the return 1 function in the  @f_onchar_delete  or @destroy trigger. Without this check, If this return is present, we get an infinite while loop.


src/game/chars/CChar.cpp
```diff
@@ -429,7 +429,7 @@ bool CChar::Delete(bool fForce)
 {
 	ADDTOCALLSTACK("CChar::Delete");
 
-	if ((NotifyDelete() == false) && !fForce)
+	if ((NotifyDelete(0) == false) && !fForce)
```
cbnolok, line 432: I'd use false here

src/game/clients/CAccount.cpp
```diff
@@ -593,8 +594,9 @@ void CAccount::DeleteChars()
 			CChar * pChar = m_Chars.GetChar(--i).CharFind();
 			if (pChar != nullptr)
 			{
+				pChar->NotifyDelete(1);
```
cbnolok, line 597: This is not needed, since it's already called by CChar::Delete, and you can pass fForce there

Jhobean:
I tested for different way and seem work correctly.

PR #1048, opened 2023-03-03
WIP: c++20 support

mtwango:
**Done** (all this changes are compatible with c++17):
- 2 external libraries updated (flat containers and zlib)
- removed deprecated std::uncaught_exception() (should affect anything, since we are using c++17, which has std::uncaught_exceptions() already)
- type cast empty string "" to CSString for consistency and error C2445 on c++20

**Missing**
- initialization error C2440 during compilation on [CServerConfig.cpp](https://github.com/Sphereserver/Source-X/blob/da1a811c00cf3601914cfa0608529cc5159f5202/src/game/CServerConfig.cpp#L700) which should be caused by `CAssocReg` and `CElementDef` constructors, but I'm not skilled enough to fix it (got it working by removing contructors).


cbnolok:
The proposal is to switch to cpp17 or 20? Is that revision compatible with latest gcc on latest stable debian version?

mtwango:
Sphereserver should already be building on cpp17 (according to [Linux-GNU_common.inc.cmake](https://github.com/Sphereserver/Source-X/blob/da1a811c00cf3601914cfa0608529cc5159f5202/src/cmake/toolchains/Linux-GNU_common.inc.cmake#L14) and [Windows-MSVC.cmake](https://github.com/Sphereserver/Source-X/blob/da1a811c00cf3601914cfa0608529cc5159f5202/src/cmake/toolchains/Windows-MSVC.cmake#L18)).

These changes provide partial support to cpp20 (see Missing, in PR) while keeping backwards compatibility with cpp17. This PR is compatible and tested with cpp17 on Windows 11 and Debian 11 both x64.

If somebody could check 'missing' part of this PR, then it would be buildable both on cpp17 and 20.

cbnolok:
I mean, which is the most recent gcc version available on the latest debian version? Does it support more of less all of the cpp20 features?
Just to know, in any case it wouldn't be a problem if compatibility is preserved for cpp17. It's important though to know if there is a wide support of cpp20 before thinking about officially switching to cpp20, in the future. In that scenario we would allow new features and break compatibility with previous cpp revisions 

Il giorno 3 mar 2023, 14:15, alle ore 14:15, mtwango ***@***.***> ha scritto:

mtwango:
Default version of gcc in Debian 11 (current) is 10 (gcc11 can be installed, but for the matter of simplicity I would assume gcc10 is default). 
Gcc10 (and even gcc11) doesn't support all features of C++20 (see [C++20 Support in GCC](https://gcc.gnu.org/projects/cxx-status.html#cxx20)) so I would think that backwards compatibility with c++17 is required.

cbnolok:
Sorry for the delay. Could you update the files with the current folder structure?
I see you have modified the flat containers library, did you test the changes? If yes, i'd gladly merge this PR.

cbnolok:
Done with last commit.

PR #1069, opened 2023-04-30
Port from 56d: missing buff icons

DavideRei:
Added to enum BUFF_ICONS missing icons from 56d.


AtillaG1:
It seems like there should be much more, but they are mixed inside the other ones feels like.
https://pastebin.com/dZYeLf6k

DavideRei:
> It seems like there should be much more, but they are mixed inside the other ones feels like. https://pastebin.com/dZYeLf6k

I added all the missing ones from 56d

image

Jhobean:
That seem correct

DavideRei:
It seems to work. It depends also on client version

image

ClassicUO doesn't show all icons
image

PR #1091, opened 2023-08-23
Monsters' AI fixed & Guild Notoriety fix

LuxionUO:
The issue with the Heuristic function in CPathFinder has been fixed. Now, the monsters correctly follow their targets without any problems. They are now able to effectively navigate around obstacles and move in a straight path. Additionally, the problem of zig-zag issue following in the South direction has been addressed.


AtillaG1:
> Added GuildNotoriety depending on Align

Just put the mystery out for the others, this notoriety thing is working for me even without this fix. I have looked through the scripts, but no @noto modifications.....
There is something weird that Im the only one it works for :octocat:

PR #1115, opened 2023-09-26
Fix char delete trigger appear for NPC

Jhobean:
Second fix for issue https://github.com/Sphereserver/Source-X/issues/1029

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3016,7 +3016,6 @@ Note: The only way the server has to know if the bankself is closed is to store 
 - Changed: @UserQuestArrowClick, using return 1 now prevents from showing default message when (left/right)clicking arrow.
 - Changed: Crafting skills now won't display the bounce message when successfully crafting an item, with the skill message is enough.
 
-
 10-08-2022, Drk84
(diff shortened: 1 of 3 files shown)


Jhobean:
Ok, All work like expected except this:
When we delete a char with the client screen, f_onchar_delete is launch 2 times. The only way to avoid this is to create a second bool argument on char delete function.
image



image

Tolokio:
Tested deeply and works fine. But I think:

-There is no need to set argo=char  as char is already in <src> and works fine in both cases (from login screen and ingame command)
-Leave argo as m_pclient (it dosnt work on any serv_trigger) or change it to be account if possible.

Found minor weird bug:
If u "remove 1" an online player and deny it by return 1 on f_onchar_delete, player will not be removed from game but from players screenshoot. SO player can't move until sphere resend his data back.

Jhobean:
Seem complete like this.
I added WARNING RED message each time a character is deleted. When character is deleted via login screen, you have a little message and WARNING message just after.
image

cbnolok:
That fixes also the bug reported above?

Jhobean:
> That fixes also the bug reported above?

Not yet, ill check

Jhobean:
BUg already fixed, ready to merge for me

Issue #1118, opened 2023-10-02
@famechange @karmachange issues


Soulless-1:
A recent commit seems to have rendered these triggers unusable, even in their prior state they were not robust but they were useful @famechange and @karmachange

https://github.com/Sphereserver/Source-X/commit/7de9bb49594f0a8274469012e625e687317c1cc6

assuming it would work like the following
                local.old = fame/karma before change
                local.new = fame/karma after change
                argn1 = amount it is changing
   
however it works like so. 
              local.old = fame/karma before change
              local.new = local.old + argn1  (old karma + new karma combined)
              argn1 = total karma is set to

you can see that the changes are not necessarily very useful. i propose a change: 
             argn1 = amount it's changing (adding/removing) (r/w)
             argn2 = the creature that triggered the change (if present)
             local.old = old value before the change
             local.new = new value after the change (r/w)

and i don't think that just setting the value on a character should set the trigger off. for a very important reason, karma and value changes are subjective based on the characters current values. Not every character gets the same amount, so if you make a proper system you will be doing formulas based off the target's fame+karma and the killers fame+karma. and if you want to just set karma on chardefs and/or characters in game you want the value set exactly as you typed it.

Issue #1117, opened 2023-09-27
f_onchar_create_init dosnt work


Tolokio:
as title said. It dosnt work.

![image](https://github.com/Sphereserver/Source-X/assets/20745578/5f27de2d-1214-4bf2-99d4-999c4916c2b3)

f_onchar_create does work

![image](https://github.com/Sphereserver/Source-X/assets/20745578/14b9b38f-c8b2-4801-a98c-4a6b63c5bb46)

I think it dosnt work cos it is called from Client and client is always 0 on any serv_trigger

Tolokio:
![image](https://github.com/Sphereserver/Source-X/assets/20745578/088d9aff-681c-42ac-bcb0-401940702888)

This check never get fired, we could remove it cos it is not needed anyway.
This is what happening if u try to create a new character 
![image](https://github.com/Sphereserver/Source-X/assets/20745578/ecbab897-7b87-432a-a5fe-685484abc2ad)

And it denys character's creation.

![image](https://github.com/Sphereserver/Source-X/assets/20745578/e10ebf5e-4818-41cf-9893-4628af5216e3)

drk84:
f_onchar_create_init doesn't work because the 2nd parameter is null, it should be pChar or &gServ. 
When you do serv.log it tries to call the command on the 2nd parameter, but because there are no object nothing is displayed (it actually trhrows a warning)

xwerswoodx:
&g_Serv is probably better choice here, while pChar is initialized after this function triggered.

Issue #1084, opened 2023-07-18
Npcs with spawn item inside buildings created outside buildings


DavideRei:
If you put a char spawn inside a building the npc can be created outside the building.
Even in a dungeon, if you put a spawn in a corridor next to other rooms/caverns, the npc can be created in other rooms/caverns.

![image](https://github.com/Sphereserver/Source-X/assets/118212274/9236ab4b-e43c-44b7-a5b5-a83c3c61486a)

You have to keep low the spawn range if you want the char to spawn where you want to.
On 0.56d (and 0.56c) is not like that, the creature spawns only inside the building/room/corridor where is the spawn item.

drk84:
HMm i doubt the spawn gems ever checked the room  shape

DavideRei:
You can try on 56d. If you are in a room the npc is created outside only if near the door and with the door open. I remember it like this since 2013.

![Screenshot_2023-07-19-00-22-10-760_com android chrome](https://github.com/Sphereserver/Source-X/assets/118212274/052ff79b-c0e9-4750-8667-95c7f9237c50)

It's not a big deal, it's just that it was convenient.

Jhobean:
X fork was in 2016. So the changelog you show is implemented on X.
When spawning, server choose a random position and verify if char is stuck.

AtillaG1:
I can confirm this is happening, thats why I had to add the check if its dungeons, you spawn NPC on spawnpoint, otherwise it can be placed over the locked door, which you can unlock by killing this monster (impossible).
Back then Ive been told the spawn checks for not spawning outside of room, or in dead ends, but it never did.

AtillaG1:
If anyone need im using alternative solution for the dungeons (r_dungeon) event needs to be on the region, until fixed. 😇

> IF (<SPAWNITEM>) && (<SPAWNITEM.REGION.ISEVENT.R_DUNGEON>) 
> 	P=<HOME>
> ENDIF

AtillaG1:
for me it seems like it works, if the NPC should be spawned outside the room you cant access, then its spawned directly on the spawn, correct?

DavideRei:
It works! Thank you!

Issue #1083, opened 2023-07-04
Faction cant be read from chardef


AtillaG1:
Faction on npcs cant be read from chardef. Im setting the faction in on@create.

![image](https://github.com/Sphereserver/Source-X/assets/25097484/609c3474-91c1-4d95-9d8c-d56e0e9c17f3)

with .xshow faction
![image](https://github.com/Sphereserver/Source-X/assets/25097484/112c8605-3774-444b-ad99-d65c2890589b)

DavideRei:
You can't read properties declared under @create with serv.chardef

I think that faction should work in chardef or the sphere x chardef scripts are wrong

xwerswoodx:
Honestly as DavideRei said you can't get variables set after char created by using chardef or itemdef etc. Because the values is not saved on anywhere as they defined after you created the npc.

The only way to get these variables ia spawning invisible new npc and get the variable and remove it. Or you can place one hidden in your star chamber and you can put it's uid to your scripts for checks. Or you can search map with forchars to get one npc with same id to take info from him.

This is not a bug at all, serv.chardef is for the properties that defined inside [chardef c_...] section. Any property defined under triggers cannot be reach with it.

PR #1120, opened 2023-10-03
Couple of Fixes and Feature Updates

xwerswoodx:
- Fixed: Re-Coded @KarmaChange and @FameChange triggers not setting variables correctly. (Issue #1118)
1. ARGN1 now return as the amount of Karma/Fame being added.
2. ARGN2 is now return oldKarma. (ARGN2 + ARGN1 == New Karma)
3. LOCAL.NEW removed, while ARGN1 is writeable, it's impossible to track it inside the trigger.
4. ARGO now return the source of karma/fame income if exists.

- Fixed: f_onchar_create_init is not called. (Issue: #1117)

- Fixed: IF and QVAL statements cannot evaluate correctly 32bit+ bitwise conditions. (Issue: #1078)

- Added: @PartyAdd trigger. (Feature Request: #1054) 
1. I: The player who is joining to the party.
2. SRC: The player who is inviting. 
3. Return 1: Cancels the action.

- Added: LOCAL.SOUND and LOCAL.ANIM under @Start (Skill) and @SkillStart triggers to override sound and animations before start crafting or gathering.

- Fixed: FACTION property not saved and load correctly under [CHARDEF] definition. (Issue: #1059 and #1083)

- Added: Added COMBAT_PARALYZE_CANSWING (080000) to combat flags to let characters continue attack while paralyzed like old behaviour. (Issue: #1124)

- Fixed: ATTACK command causes the player getting criminal without checking AttackingIsACrime value in sphere.ini (Issue: #1123)

- Added: Added CAN_I_EQUIPONCAST (08000000) flag for items to let scripters customize allowed items while EquippedCast disabled.

- Fixed: Spells can trigger @Success triggers and play animations if they paralyzed after selected skill.

- Fixed: Motivation < 0 causes STATF_WAR flag stay infinitely on characters. (Issue: #517)

- Fixed: The issue where characters can spawn out of the rooms. (Issue: #1084)

- Fixed: More1 and More2 values readn as integer instead of dword (unsigned integer) (Issue: #784)


cbnolok:
Aren't karma/famechanged redundant triggers? In karmachange i could get the old value with <karma> and the new with karma + argn1

xwerswoodx:
> Aren't karma/famechanged redundant triggers? In karmachange i could get the old value with and the new with karma + argn1

The problem is argn1 is writeable, and it can be changed after on_trigger. And setting local after karmachange triggered not affect local.new value. Because we have to send arguments before argn1 changed so after argn1 changed it's impossible to update Local.new for trigger.

For example, if you .set karma 100 and your karma is 50, you get old=50 new=100 and argn1=50 its fine but when you use argn1=200 under trigger, it returns old=50 new=100 and argn1=200 because argn1 updated after on_trigger but we have to send args inside on_trigger function so to update args we have to send again but this can cause endless loop as player have argn1=200 inside trigger so it keep try to trigger endlessly to update everytime. Or I don't know if there is a way to update local variables after trigret returned.

cbnolok:
I mean, do we really need a double trigger and a local.old/new when we could just do karma + argn1? Or, we could set argn1 = current karma (read-only), argn2 = karma change.

xwerswoodx:
> I mean, do we really need a double trigger and a local.old/new when we could just do karma + argn1? Or, we could set argn1 = current karma (read-only), argn2 = karma change.

Honestly for me it's totally useless, even we didn't have local.old and new before and I have never look for it, because as you said, we can get these values by functions easily. But they added local.old and new, and as they accepted, I didn't want to remove it directly because may someone use it like Soulless using them.

For me, while src.karma already giving old karma and argn1 should give karma change (which is broken in current build) and karma+argn1 should be enough to track all values, but I don't really know why local.new and old added at all. I read the change log and I made a change to make it how they should be, but removing them is not a big deal for me, as I said I just keep them because they already added.

cbnolok:
I get your point, but i'd rather remove them than adding a new trigger :D

xwerswoodx:
> I get your point, but i'd rather remove them than adding a new trigger :D

I removed new trigger and send local.old in argn2 directly if anyone wish to use :))

src/game/items/CItem.cpp
```diff
@@ -2247,12 +2247,6 @@ void CItem::r_WriteMore1(CSString & sVal)
     ADDTOCALLSTACK("CItem::r_WriteMore1");
     // do special processing to represent this.
 
-	if (IsTypeSpellbook())
-	{
-		sVal.FormatHex(m_itNormal.m_more1);
```
cbnolok, line 2252: Can i ask why was this and the similar function for more2 removed? For spellbooks, more1 and more2 are bitmasks holding the known spells, it's more readable as hex than as decimal

cbnolok: I mean, there's FormatULLHex

xwerswoodx: because it's already done it automatically in this case, there is a Switch under the check and default return is sVal.FormatHex(m_itNormal.m_more1); so second check for same thing seems unnecessary for me

        default:
            if (CResourceIDBase::IsValidResource(m_itNormal.m_more1))
                sVal = g_Cfg.ResourceGetName(CResourceID(m_itNormal.m_more1, 0));
            else
                sVal.FormatHex(m_itNormal.m_more1);
            return;

these are the lines I mentioned.

src/game/CWorldImport.cpp
```diff
@@ -471,12 +471,12 @@ bool CImportFile::ImportWSC( CScript & s, word wModeFlags )
...
 			else if ( s.IsKey("MORE" ))
 			{
-				pItem->m_itNormal.m_more1 = atoi(pArg);
+				pItem->m_itNormal.m_more1 = (dword)atoi(pArg); //Need to read as dword not an int.
 				continue;
 			}
 			else if ( s.IsKey("MORE2" ))
 			{
-				pItem->m_itNormal.m_more2 = atoi(pArg);
+				pItem->m_itNormal.m_more2 = (dword)atoi(pArg); //Need to read as dword not an int.
```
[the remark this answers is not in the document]
xwerswoodx: I haven't used str_toui before but I can pr in a minute after testing it :)

src/game/items/CItem.cpp
cbnolok: Oh i see, good one!

src/game/CWorldImport.cpp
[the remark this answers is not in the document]
xwerswoodx: Updated, I haven't know there is a str_toUI ULL functions, I should check strings file more often :))

Tolokio:
Ill try to find time this weekend to test everything. nice work.

Issue #1112, opened 2023-09-26
Lumberjacking doesn't work with non file static trees


DavideRei:
.add i_tree_5

Equip an hatchet and try to chop the tree: "you cannot chop so close to yourself".
If i add SKF_NOMINDIST to the skill flags i get "try chopping elsewhere".
No worldgem bit is created

It's the same with trees with static flag saved in spherestatics.scp

drk84:
Woah i have noticed even normal trees doesn't produce any wood

Jhobean:
Some repport on my shard about this too. Player try different tree far from city and sometime it work. Weird.

xwerswoodx:
Was a really simple issue but hard to notice 👎

Jhobean:
Ho god!!!  Nice fix it was a weird intermittent bug

DavideRei:
It works! Thank you

Issue #983, opened 2022-12-11
Unequal Z betwen uoclient and sphere


Tolokio:
This issue is related to #617 and #877

There is a problem with inclinated tiles. As soon players step on a inclinated tile, Sphere and UOClient will give diferent Z to players.
Player will not realize until client is refreshed, for example if player is affected by "flip", "resend", "update" or "fix". As soon refresh happens, sphere will send the real Z to uoclient.
This use to happen when people fish on a cliff, character is flipped and new z is sent. Moving the player up or down across Z axis.

This issue can cause other issues related to height of the player. There can be places where player will not be able to step on even if client allow it.  In certain cases it may cause canseelos fails.

Depending on the tile inclination and the tile where u come from, the real player's z can be higgher or lower than the one UOclient is using.

Note: Player's Z at sphere and Uoclient are already unequal before the refresh. At the refresh sphere just tell the client where the player realy is and client update the Z.

Tolokio:
![image](https://user-images.githubusercontent.com/20745578/207864799-983ef153-83e2-4211-a556-583082246d78.png)
I think problem got to be at this functions. there are 4 functions like this.

Jhobean:
A new method for determining the reel Z should be create and place somewhere.

Example 
servuo: https://github.com/ServUO/ServUO/blob/05cdc8780ecd81a6abca0f5f14097c8e6bc4860f/Server/Map.cs#L550
Classic UO: ?

ORION:
```
void CLandObject::UpdateZ(const int32_t &zTop, const int32_t &zRight, const int32_t &zBottom)
{
	if (m_IsStretched)
	{
		//Сохраним среднее значение Z-координаты
		m_Serial = ((m_Z + zTop + zRight + zBottom) / 4);

		//Значения для рендера
		m_Rect.left = m_Z * 4 + 1;
		m_Rect.top = zTop * 4;
		m_Rect.right = zRight * 4;
		m_Rect.bottom = zBottom * 4 + 1;

		if (abs((int32_t)m_Z - zRight) <= abs(zBottom - zTop))
			m_AverageZ = ((int32_t)m_Z + zRight) >> 1;
		else
			m_AverageZ = (zBottom + zTop) >> 1;

		//Минимальная Z-координата из всех
		m_MinZ = m_Z;

		if (zTop < m_MinZ)
			m_MinZ = zTop;

		if (zRight < m_MinZ)
			m_MinZ = zRight;

		if (zBottom < m_MinZ)
[...]

Issue #877, opened 2022-07-24
Height Issues


Soulless-1:
Looks like sphere has some height issues, which cause a few problems with preventing players moving around. There are quite a few spots, but here are two examples of what the issue is :

in this example you cannot walk in these areas if your height is over 12, because the roof would be "too low" for you, however this is not a place that players are restricted from walking in UO, and in this tight area it impedes a lot of players pvping in this popular building 
![image](https://user-images.githubusercontent.com/17236327/180660494-658648a4-7172-48a5-bd39-1a7cb7bfa741.png)

and if this example if you are under 15 height you cannot walk up this hill
![image](https://user-images.githubusercontent.com/17236327/180659610-1482fe19-3301-4f1d-8f91-8318a55dba12.png)

there is no setting you can have in the current system that will not result in issues for players, and i do not see a way to turn off this system in the sphere.ini

You are short enough to fit under all buildings and too short to walk up some hills
OR
 you are too tall to fit under some rooves but tall enough to walk up all hills

Tolokio:
the outside from britain sewers have use to need a ladder or something from many years if i remember right.

But there are certain height issues. I did a post about it.
Duper items takes height from original item, So many heights are bad.

Tolokio:
@Soulless-1  Could u add more info abot the bbuilding issue. U said client allow to walk buy sphere denys it.
can u tell the Z of the floor and roof? and do .xshow height of both as well?

For the second, should we add a climbing trigger, or at least we should dicuss abaut whhat would be the max Z a char should b able to climb

Soulless-1:
No, when I reference UO i mean the game ultima online traditionally does not prevent the player from going into buildings due to height. In my examples those are the types of things players should be able to do under all circumstances, otherwise it limits them access to parts of the map, or may cause them to get stuck in areas.

this isnt about item height or anything like that. 
the climb trigger would be separate from this, it's about being able to walk.

Soulless-1:
working on a test build, will close out when PR complete, thank you Xwerswoodx

cbnolok:
Any news on this?

Soulless-1:
indeed fixed, need a height of 12, but i can walk in all areas. thank you!

Issue #779, opened 2021-10-06
Spell @success is triggered for Teleport even if the spell was blocked 


a1exp:
When you try to cast Teleport to a fence for instance you can't do it and the following sysmessage is shown. However spell @Success trigger still fires. 

![image](https://user-images.githubusercontent.com/31130012/136208735-cb007147-8cd6-4f44-8db7-18686d8feac3.png)

It breaks some logic in my scripts, e.g. I want to show an additional animation upon successfull teleporting. 

Is there a way to trigger @fail in such cases? Or are there any other workarounds?

a1exp:
Hm.. In my case the sysmessage "You can't teleport to that spot." is printed, but the spell does not fizzle when I try to cast to a fense (statics) with a player. Will double check with with default script pack.

Jhobean:
@a1exp yes you are right. I just forgive to mention on my screen shot I modified the code...  But the fizzle behabior is bad.. what do you think?

a1exp:
@Jhobean 

yep, just reproduced with default script pack.

ideally, the check if the player can teleport to a given spot should be done on spell @select / @start and do not trigger @success. But I don't not if it's doable code wise. 

Triggering @fail and fizzle should be also acceptable, what is you concern? 

By the way, I checked Recall spell for a similar situation (marked a rune in anti-magic zone by GM, gave the rune to a player). When the player casts Recall, the spell fizzles and two sysmessages are displayed: 
 - `The anti-magic field blocks you`
 -  `You stop casting the spell.`

But the weird thing is that BOTH @fail and @success are triggered in such case. So the @success issues seems to be common for both Teleport and Recall spells. 

![image](https://user-images.githubusercontent.com/31130012/136674464-85084f1e-c184-4456-a551-2bd535d82fb8.png)

xwerswoodx:
I put that as high priority take a look at it tomorrow

PR #1128, opened 2023-10-11
Z Correction Tests

xwerswoodx:


Tolokio:
Tested with 7.0.97.25
-It is hard to find a spot where sphere gets different Z than UOClient. In thoose cases, diference is -1 or +1, much better than before.
-This fix many terrain issues where Uoclient allow u to climb a hill but sphere dosnt. It is almost impossible to find one of this cases now.
-I just find a case where heights get mess but I think it is very rare case. I upload a ScreenShot
image
At screenshoot u can see diference betwen sphere(blue) and uoclient(orange) heights.

The arrangement is not perfect but offers a substantially significant improvement

xwerswoodx:
I push another build, it should be fixed for the location on the picture, I am still doing a check if it causes any other issue

Tolokio:
Much better fix,  Xwerswoodx found two spots with +1 and -1. I wasnt able to find any. For me is enought accurate to be implemented. Thoose spots Xwerswoodx found are just very very rare cases and just +1 or -1. We should check if runuo or servuo have the same problem there just for curiosity.

Tolokio:
I ended my method to fix height, I think it does not worth at least how I did it. I think it increase cpu and memory usage a lot.

xwerswoodx:
Yeah a bit silly, I would like to do some more checks to make it more accurate but it uses so much cpu if I do it, so I think with this way it can be best to stop, I am not sure what else I can do anymore honestly.

Jhobean:
(Issue: #983, #617, #877) (Issue: #779)(Issue: #1112) (Issue: #720)

PR #1178, opened 2023-11-15
Editable DispID for chars 

DavideRei:
Editable DispID for chars to change char graphic/body without changing baseid/defname, like it was in 0.56c


Jhobean:
I tested it. Work Great. Cool feature

Raylde:
LGTM

cbnolok:
There's a problem with that. You are changing the CHARDEF (CCharBase) dispid, not the dispid of a specific instance of that char (CChar).
I'd do the following: create a CChar::m_dwDispID member and a CChar::SetDispID method to set that. If a valid dispid wasn't set (0 or CREID_INVALID), get the chardef one with Char_GetDef() and return that

Jhobean:
> There's a problem with that. You are changing the CHARDEF (CCharBase) dispid, not the dispid of a specific instance of that char (CChar). I'd do the following: create a CChar::m_dwDispID member and a CChar::SetDispID method to set that. If a valid dispid wasn't set (0 or CREID_INVALID), get the chardef one with Char_GetDef() and return that

Are you saying if I change a specific zombie to c_bird, all zombie will look like bird?

cbnolok:
From the code, it looks so

DavideRei:
lol it's true, just tried.

DavideRei:
Fixed

DavideRei:
it seems that if i set dispid on a npc with a spawn, the dispid remain attached to the spawn. Even if the npc dies, the modified graphic remain on the new spawned npc. The new dispid seems to remain saved somewhere.
How is that possible? @cbnolok

Jhobean:
Need to do a manual .update to see modification. possible to force it in the server?

Jhobean:
I'm not able reproduce the spawn point bug you mentionned. Seem ok on my side

DavideRei:
yes, seems fine now

Jhobean (approved):

Issue #1142, opened 2023-10-24
[Change] Snooping distance can be adjusted with the skill range value.


canerksk:
CCharSkill.cpp found in

```
if (GetTopDist3D(pCharMark) > 1)
{
	SysMessageDefault(DEFMSG_SNOOPING_REACH);
	return (-SKTRIG_QTY);
}
```

It is set as 1 frame in the line, but this may vary depending on the servers or it may be requested to be retroactive, in the past it was 2 frames. This can be adjusted based on Skill Range value

eg;

```
	CSkillDef *pSkillDef = g_Cfg.GetSkillDef(SKILL_SNOOPING);
	int iMaxRange = pSkillDef->m_Range ? pSkillDef->m_Range : 1;
`	if (GetTopDist3D(pCharMark) > iMaxRange)
	{
		SysMessageDefault(DEFMSG_SNOOPING_REACH);
		return (-SKTRIG_QTY);
	}
```

_The codes here are purely examples. Although many have been tested, some may need to be rewritten._

Jhobean:
Good Idea. Apparently you have the script. Are you ablento push a PR?
Important behavior should be to have default 1 if you write nothing

mtwango:
This should already be resolved in #1338

PR #1183, opened 2023-11-19
Feature/snooping use range

raydienull:
Implement #1142 

and minor code clean up of unused vars


src/game/clients/CClientEvent.cpp
```diff
@@ -422,9 +421,6 @@ void CClient::Event_Item_Drop( CUID uidItem, CPointMap pt, CUID uidOn, uchar gri
 			}
 		}
 
-		if ( pOldCont != pItem->GetContainer() )
```
cbnolok, line 425: I remember adding this for a reason, does git blame help for that?

Raylde: i blame it and see that doesnt change in years
https://github.com/Sphereserver/Source-X/blame/ae696679c35edce7192b8fbb55350ba28003718d/src/graysvr/CClientEvent.cpp

xwerswoodx: This check should be there, because this check if the item changed the container under dropon_item trigger, I mean think like 2 players tries to get same item, and you did some checks under dropon_item, for the first player, then added the item in his container, you don't want to second player take it away from player's backpack. This check is added to block this, but it probably could be really really rare issue, as sphere can handle these checks in milliseconds, but as someone can have tons of checks under @DropOn_Item, checking if the old container is same should be guarantee. So it has to stay like it was honestly.

Jhobean:
Duplication of https://github.com/Sphereserver/Source-X/pull/1338?

PR #1184, opened 2023-11-19
Feature/meditation fail movement

Raylde:
Add new feature to enable/disabled if meditation fail on movement


src/game/CServerConfig.cpp
```diff
@@ -196,6 +196,7 @@ CServerConfig::CServerConfig()
 	_uiExperimentalFlags= 0;
 	_uiOptionFlags		= (OF_Command_Sysmsgs|OF_NoHouseMuteSpeech);
 	_uiAreaFlags		= AREAF_RoomInheritsFlags;
+	_bMeditationMovementAbort = false;
```
cbnolok, line 199: Not that i understand really the reason, but the prefix for bool is f in this codebase, could we use the same one?

PR #1194, opened 2023-12-11
Fix FUNC template when create via ingame command

raydienull:


cbnolok:
Is this tested? The code seems legit

Issue #1097, opened 2023-09-06
CAN_O_NOSLEEP and sphere reboot bug 


AtillaG1:
hi there,

this one seems like its working now. #1082
Now there is a problem, if you reboot sphere, all the items and sectors are set to sleeping.
This means the items with flag can_o_nosleep sleeps immediately and never tick.

xwerswoodx:
I can confirm that items with CAN_O_NOSLEEP go sleep after server reboot.

AtillaG1:
Heyo, any chance anyone had time to look at this thingy? 🦕

PR #1198, opened 2023-12-12
Fix issue #1097 (CAN_O_NOSLEEP not work on server reboot)

Jhobean:


AtillaG1:
tested and works for me. 🤩

Raylde:
LGTM

PR #1200, opened 2023-12-14
Fix Attack stop #684 

### LINKED ISSUE #684: Attack stop
if you attack and cast spell the fight is canceled for your char, the other stay on fight
tested in pvp

raydienull:
Fix for make player still attacking when dont have skill active but s…tay on war and have Fight target

Close #684

### DIFF
--- a/src/game/chars/CCharAct.cpp
+++ b/src/game/chars/CCharAct.cpp
@@ -4578,6 +4578,18 @@ bool CChar::_OnTick()
             }
         }
     }
+	else if (m_pPlayer)
+	{
+		if (Skill_GetActive() == SKILL_NONE)
+		{
+			// if player are in war and have target to attack
+			if (IsStatFlag(STATF_WAR) && m_Fight_Targ_UID != UID_UNUSED  )
+			{
+				//Start fight skill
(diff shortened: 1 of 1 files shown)


AtillaG1:
I think this one can be pushed too, if agreed by devs, no?

Tolokio:
I dont want to stop the pull or disturb anyone but I need to understand this and u know how I am... I can't test with this pc.

Does this affect to any skill or just magery? For example, healing yourself also stop you from attacking? I see noone saying "bandages stop combat"

image

Following drk's logic it should stop attacking.

I dont like the fix cos if I leave my player just doing nothing or just walking around, that check will be firing once per second (not too much but we could avoid it doing it only when neccesary. If it is related just to magery, we can do it just when magery ends.
And if it only related to magery that means there is a bug somewhere and this is not a real fix cos bug is not solved.

If it is happening with any skill as drk84 suggest. We can put the fix when a skill is done or when action is changed to: "skill_none". No reason to let this fix be firing again and again once per second. I repeat is it not too much but you know.... "performance"...

If we are sure this is the best option for me is fine. But lets be sure.

PR #1201, opened 2023-12-15
Change LAYER_AUCTION to LAYER_STABLE and now is used to store the sta…

raydienull:
…ble pets in the player stable layer

### DIFF
--- a/src/game/chars/CCharNPCAct_Vendor.cpp
+++ b/src/game/chars/CCharNPCAct_Vendor.cpp
@@ -105,15 +105,6 @@ bool CChar::NPC_StablePetSelect( CChar * pCharPlayer )
 	if ( ! pCharPlayer->IsClientActive())
 		return false;
 
-	// Might have too many pets already ?
-	int iCount = 0;
-	CItemContainer * pBank = GetBank();
-	if ( pBank->GetContentCount() >= g_Cfg.m_iContainerMaxItems )
-	{
-		Speak( g_Cfg.GetDefaultMsg( DEFMSG_NPC_STABLEMASTER_FULL ) );
-		return false;
-	}
-
 	// Calculate the max limit of pets that the NPC can hold for the player
 	double iSkillTaming = pCharPlayer->Skill_GetAdjusted(SKILL_TAMING);
 	double iSkillAnimalLore = pCharPlayer->Skill_GetAdjusted(SKILL_ANIMALLORE);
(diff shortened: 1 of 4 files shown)


Tolokio:
I use layer auction for other things... Are u removing it?

Jhobean:
He change the name to stable and chage behavior of stablemaster storing pet there.  You vam use this layer again with no problem.

cbnolok:
Since it can be used in multiple ways, while holding stabled pets is done only by one specific npc kind, what about giving it a more generic name? To remark it could be used for other stuff

Tolokio:
I dont know if I understand well what u say. If players store pets there, Could it be used for others things? I mean, Are u saying it  can store pets and more stuff? 

nolok said "stabled pets is done only by one specific npc kind", I dont understand. It is an npc that uses that layer, but it uses player's layer, not vendor layer, right?

Jhobean:
Layer can be use for what you want. Just use the layer number like you did previously. Name is just a define in the source. 

But be aware, this layer will now receive you pet stabled.


That said, now npc stablekeeper now use this layer ON the player to store pet.before npc was storing pet on the NPC bank because If you store pet on bank of player, player was able to see it. With this fix, pet will be now store on player on this layer. Before this fix, If you kill a stable npc, player loose all the pet the NPC had on his bank.

@Raylde i say true?

cbnolok:
I mean that the layer can be used for auctioned items, stabled pets and other stuff, so instead of changing a specific name for another specific name, i feel that a more generic one would be more appropriate

Tolokio:
OK, I took a look on the source. Layer can be used for any kind of stuff as stable system only count t_figurines. Anyway this change have 2 problems or possible problems for servers that use auction system. 

1: If someone use auction system to try to sell a horse (t_figurine) or any other mount using that layer, that mount will appear on the stable system. (unless t_figurine's link is not player's uid)
2: If someone use auction system, it will show stabled pets until you modify the script to ignore t_figurines. (or to ignore linked t_figurines)

//
All t_figurines are linked to the one who shrink it? or it is linked only when it is stabled?

image

As u can see, the code check if "t_figurine->link" is the player using stable system. That check was neccesary when all t_figurines  from diferent players were in the same layer of the npc. Now seems to be useless so we could remove that check. The only reason to keep that check is to solve the problems I mentioned. This way u could remove "link" from T_figurine to avoid it to be shown at stable's dialog. And u can tell auction system to ignore t_figurines that are linked to a player.

Another question: if I put a t_figurine inside a backpack that is inside layer_stable, will it be shown? or system do not check inside containers of that layer?

Jhobean:
What about keep layer auction and create a new layer for layer stable? For me its important to store the pet on the player itself. It's a major bug in sphere when you kill the npc the mount dissapear lol.

Tolokio:
> What about keep layer auction and create a new layer for layer stable? For me its important to store the pet on the player itself. It's a major bug in sphere when you kill the npc the mount dissapear lol.

There are few ways to avoid ur problem by script.

1-
Make Invul your stable master.
2-
ON=@death
 f_get_mounts_into_players_bank
 3-
 on=@death
 f_get_mount_somewhere_and_get_them_back_to_stable_master_when_he_spawn.
  //u can use the spawn to store the id of the container where the mounts are. And call it back when a new stable_master spawns. It is not hard to script and for me is the best solution to your problem.
 
 
 
 //
 I would agree maybe creating another layer for it could be fine.  It would avoid conflict with auction_system or other custom system that use auction_layer. But if someone wants to have old system, where pets can only be retrieved where they were stabled, the options are: Create custom stable system or compile custom sphereX.
 
  Should we add a triggers to let people customize?
  
 On=@petretrieve
 local.figurine -> uid of the figurine inside the for
 Return 1 ->Avoid pet to be shown at stable's dialog.
 
 On=@petstable
 src->Player
 |-> Stable_master npc
 argo->Pets trying to be stabled
 
 
Whatever u choose to do is fine for me. Im forced to compile my own sphere anyway and I dont care to much about this. But I like to share my point of view as it seems that nobody thought about the possible problem or backwards compatibility.

PR #1205, opened 2024-01-04
CSFile Windows creation file mode fix

DavideRei:
The correct mode to create a file in Windows, if not exists, and open it, if exists, is CREATE_ALWAYS. With (OPEN_ALWAYS | CREATE_NEW) the mode is set to 5 (TRUNCATE_EXISTING) and if the file doesn't exists the CreateFile function return an error. https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea


DavideRei:
added creation file fix for Linux.
Create a new file request the file permissions

DavideRei:
added fix for #1206 
When changing the body/id of a char update the dispid

Jhobean:
Tested the dispid Part and now it work fine! Thx. ABout the CSfile, I don't know

PR #1207, opened 2024-01-14
Add ini switch for removing emote when eating. Now send object pointer to Eat trigger 

Jhobean:
1- Add an ini switch for emote.
2- Possibility to change hardcode text
3- Added Argo parameter on @eat trigger


src/game/chars/CCharAct.cpp
```diff
@@ -2238,9 +2238,12 @@ void CChar::EatAnim( lpctstr pszName, ushort uiQty )
 	if ( !IsStatFlag(STATF_ONHORSE) )
 		UpdateAnimate(ANIM_EAT);
 
-	tchar * pszMsg = Str_GetTemp();
-	snprintf(pszMsg, Str_TempLength(), g_Cfg.GetDefaultMsg(DEFMSG_MSG_EATSOME), pszName);
-	Emote(pszMsg);
+    if (!(g_Cfg.m_iEmoteFlags & EMOTEF_EAT))
+    {
+ 	    tchar * pszMsg = Str_GetTemp();
+	    snprintf(pszMsg, Str_TempLength(), g_Cfg.GetDefaultMsg(DEFMSG_MSG_EATSOME), pItem->GetName());
```
cbnolok, line 2244: Could you add a check to ensure pitem is not a null pointer?

Jhobean: I added a check in this particular situation but it's requiere everywhere I use pItem?

cbnolok: More checks are always better: if you dereference a null or invalid ptr, sphere crashes. If you are assuming a ptr is always valid and not null, it's best to use an ASSERT statement, both to comunicate your assumption and catch earliest possible violations of that principle.

Jhobean: OK. I place an assert on begining of function. ALl the place this function is use, pItem is here. 

I removed the null PTR check before pItem utilisation because Assert will flag if pItem is null. (THis situation should never happen but in case we will have a flag)

Issue #1215, opened 2024-02-17
Follower Commands Don't Work.


Nefrelius:
There is no need to elaborate on the problem here.
We know that there are commands for followers such as kill, come, stay and these are heard by the follower closest to us. But when the player types ALL <command>, all the followers around him should obey the command. But there is no such thing :) Even if you type Kill or All Kill, only the follower closest to you hears the command.  BUT! The player once typed ALL KILL and the closest follower attacked. Then if the player types ALL KILL again, all followers attack. I think there is a small query error.

[translate DeepL]

xwerswoodx:
It looks like working for me , I spawn 5-6 horses around and when I type ALL KILL all of them start attacking, are you sure it's not a script-side bug? Could you reproduce it with an clean sphere?

Soulless-1:
i tried it also myself. as non gm, surrounded by many untamed animals and my tamed animals. they did attack the target
![image](https://github.com/Sphereserver/Source-X/assets/17236327/c161bb60-d124-45cf-8e60-3b9228ad0e5c)

in the past though i did report some issues, and i believe they were resolved, at least it appears that way as of current.

DavideRei:
There is a bug in CChar::NPC_Act_Follow, it the pet follow his master, it uses m_Act_UID else it uses m_Fight_Targ_UID even if the pet is not in combat!
I can fix this

xwerswoodx:
I tried again with the info you gave, I used all kill, attacked all npcs to the one target, then used all come and they start following me, used that way many times even tried to change all kill target, they always listen what I told and changed their target, even stop and following me etc, but all command range looks a bit small maybe this makes player to think it doesn't working, I will check the range

AtillaG1:
Thats is great to hear!!! I will comment my custom commands and try it again!
Also "All guard me" should work too right? automatically attacking what im attacking.

xwerswoodx:
Yes, I tested all guard me and it works fine too, as soon as I tried to attack, every npcs around me start attacking it

PR #1220, opened 2024-03-19
Block spelleffect on ridden chars and other fixes

DavideRei:
If you place multiple fire field spell on the ground, go over it while riding an horse and .serv.save @spelleffect is triggered on the horse. To solve this problem i blocked spelleffect and gethit blocked on ridden chars. Don't know why during saves ridden chars could be affected but i think that while ridden they shouldn't be get hitted by spell or damage.
Fix duration of hallucination spell.
Fix wrong layer for potion delay


Tolokio:
The mount fix should be done by script while we find whats going on.
In my case I need to be able to damage mounts.

DavideRei:
The problem is that the offline mount is teleported on the p of the owner when you save.
I don't understand why an offline mount should take damage, especially from magic fields.
@Tolokio can i ask why do you need mount damage? for DOT?

Jhobean:
ON 55i mounted mount was teleported to 1,1 Now on X, mount follow you but on save I think they are place directly on you that why save create this bug.

Jhobean:
> The mount fix should be done by script while we find whats going on. In my case I need to be able to damage mounts.

Do you have a script fix to propose? We could add it to script pack.
We could add it on e_horses

DavideRei:
We could block spelleffect on ridden char if the spell has spellflag_field or spellflag_area

Jhobean:
> We could block spelleffect on ridden char if the spell has spellflag_field or spellflag_area

This option seem interesting. That mean if pet is poisoned, it will take damage.

xwerswoodx:
Honestly I am not sure about blocking every hit while Horse is riding, because if it's poisoned, it has to get damage from the poison until die or poison removed, but I agree about field, we may found better solutions for it because I think any ticking damage like poison shouldn't be blocked when player ride the mount.

DavideRei:
I removed the block on getHit for ridden chars and blocked the spelleffect on ridden chars only for spells with SPELLFLAG_FIELD or SPELLFLAG_AREA.
Also i changed NpcActFollow to take the fight target only if the npc is in combat, maybe is related to #1215

I don't why github put in the pull request old commits

Jhobean:
I think you roolback some PR 🤣  you should merge master ti your branch or redo your PR from a clean branch

PR #1252, opened 2024-06-18
Followerslots with EF_FollowerList - FIX #1251

### LINKED ISSUE #1251: ON=@FollowersUpdate not working as intended?
> Added @FollowersUpdate trigger called in any update of followers related to OF_PetSlots:
> 	argn1 (R/W): 0 = Adding followers (taming, unshrinking, summoning, buying ...), 1 = sustracting followers (release, transfer ...)
> 	argn2 (R/W): FollowerSlots used.
> 	SRC: Pet causing the trigger to run, when buying in vendors there's no PET yet, so SRC will point to the vendor.
> 	Return 1 will not allow following actions to occur: Transfer, Summon, UnShrinks from t_figurine, Buy from vendors. Any other actions (like release, set owner...) will not take in count returns.

Im using EF_FollowerList and OF_PetSlots. Im trying to use my own followerslots, since the default behaviour is not using more than 1.
Somehow, even tho i set ARGN2 to whatever amount, there is always only 1 slot used.

Is this suppose to be still linked to the maxfollowers and curfollowers, and slots? I feel like its working like half a way 😥

AtillaG1:
Fix for the review

Tested on local and seems good.

- checks for curfollowers + followerslots <= maxfollowers
- add tamed follower to the list * followerslots (2 followerslots = 2 positions in list)
- no need for extra code in @FollowersUpdate

**Issue:**
https://github.com/Sphereserver/Source-X/issues/1251

### DIFF
--- a/src/game/chars/CCharUse.cpp
+++ b/src/game/chars/CCharUse.cpp
@@ -1209,10 +1209,18 @@ bool CChar::FollowersUpdate( CChar * pChar, short iFollowerSlots, bool fCheckOnl
                 ++it;
             }
 
-            if (!fExists && ((short)(m_followers.size()) < iMaxFollower || IsPriv(PRIV_GM)))
+            if (!fExists && ((short)(m_followers.size()) + iFollowerSlots <= iMaxFollower || IsPriv(PRIV_GM)))
             {
                 if (!fCheckOnly)
-                    m_followers.emplace_back(pChar->GetUID());
+                {
+                    for (short i = 0; i < iFollowerSlots; ++i)
+                    {
+                        m_followers.emplace_back(pChar->GetUID());
+                    }
+                    short iCurFollower = (short)(GetDefNum("CURFOLLOWER", true));
+                    iCurFollower += iFollowerSlots;
+                    SetDefNum("CURFOLLOWER", maximum(iCurFollower, 0));
+                }
             }
             else
                 return false;


src/game/chars/CCharUse.cpp
```diff
@@ -1209,10 +1209,18 @@ bool CChar::FollowersUpdate( CChar * pChar, short iFollowerSlots, bool fCheckOnl
...
 
-            if (!fExists && ((short)(m_followers.size()) < iMaxFollower || IsPriv(PRIV_GM)))
+            if (!fExists && ((short)(m_followers.size()) + iFollowerSlots <= iMaxFollower || IsPriv(PRIV_GM)))
             {
                 if (!fCheckOnly)
-                    m_followers.emplace_back(pChar->GetUID());
+                {
+                    for (short i = 0; i < iFollowerSlots; ++i)
+                    {
+                        m_followers.emplace_back(pChar->GetUID());
```
cbnolok, line 1218: We could do that better, by making m_followers a vector of std::pair<ushort, CUID>, where the first number is the followerslots, value the second the UID.

cbnolok (changes requested):

src/game/chars/CCharUse.cpp
AtillaG1: I can try, but I think Im not capable of doing that. Wouldnt that need more complex changes in the whole code?

AtillaG1: I suppose its defined as CUID already 
`std::vector<CUID> m_followers;` which means we would have to review every code where is it used?

AtillaG1: Im sorry im not able to do that, I guess if we change the def of the vector, we need to change every part where the FollowersUpdate is called/used, and even with the AI help, I couldnt make it work. 😥
Anyone ideas or more experience?

cbnolok: I will take care of it, we'll need to change other code to make it working but it will be worth it.

Issue #1255, opened 2024-06-29
spellflag_no_elementalengine & SPELLFLAG_FIELD_RANDOMDECAY


GladieUO:
```
#define SPELLFLAG_FIELD_RANDOMDECAY     0x100000000
#define SPELLFLAG_NO_ELEMENTALENGINE    0x200000000
```

It seems like it does not work in the spell flags.
Dunno if its issue of too long number, or in code?

Example

```
if (IsSetCombatFlags(COMBAT_ELEMENTAL_ENGINE) && !pSpellDef->IsSpellType(SPELLFLAG_NO_ELEMENTALENGINE))

FLAGS=spellflag_dir_anim|spellflag_targ_char|spellflag_fx_targ|spellflag_good|spellflag_no_elementalengine
```

```
if (pSpellDef->IsSpellType(SPELLFLAG_FIELD_RANDOMDECAY))

FLAGS=spellflag_dir_anim|spellflag_targ_xyz|spellflag_harm|spellflag_damage|spellflag_fx_targ|spellflag_field|spellflag_nounparalyze|SPELLFLAG_FIELD_RANDOMDECAY
```

PR #1256, opened 2024-06-29
Spellflags fix reorganization #1255

AtillaG1:
Regarding this issue, #1255 I think the issue is the length of the flags, I looked through the other flags and seems like there has been free 2 spaces.
I moved the two new ones there and now they work.

Thoughts?

### DIFF
--- a/src/game/chars/CCharUse.cpp
+++ b/src/game/chars/CCharUse.cpp
@@ -1209,18 +1209,10 @@ bool CChar::FollowersUpdate( CChar * pChar, short iFollowerSlots, bool fCheckOnl
                 ++it;
             }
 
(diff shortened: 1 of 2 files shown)


cbnolok:
Which def do you have in your scripts for no_elementalengine and randomdecay? Because the code in the source seems legit and for some reason those flags aren't yet been added in the x scriptpack

AtillaG1:
> Which def do you have in your scripts for no_elementalengine and randomdecay? Because the code in the source seems legit and for some reason those flags aren't yet been added in the x scriptpack

It was added by xwerox by one of the monstrous PRs 😁 We probably forgot to add it in Xpack but I added it to mine scripts from changelog. I assumed its working, but few days ago find out sphere doest see it if set.
image
image

AtillaG1:
Removed followerslots fix from this PR.

Issue #963, opened 2022-11-05
Negative stat modifier applied on stat with lowest value, returns maximum value.


kikeking1:
When I apply a negative modifier to a stat and it exceeds the amount of the stat, instead of being 0, it deducts from 65535, causing the stat to increase.

Eg A character with 10 dex suffers a -20 penalty to his moddex and his dexter becomes 65526.

The stat modifier should be able to be negative (for example -80), but if the stat is 20, it shouldn't go above 0.

![imagen](https://user-images.githubusercontent.com/4921179/200124857-883c78e9-4627-42e5-b72c-292b750608ea.png)

Jhobean:
Similar fix with stam/hits/mana
https://github.com/Sphereserver/Source-X/pull/625

drk84:
It's not a similar fix, the difference is that modstats can go below 0 while the related stats no.

This bring the following issue:

ON=@Equip
src.modstr -50
ON=@Unequip
src.modtr 50

Currently a character with 40 str equipping this item will go below 0 and thus it will get a 65k value of str.

If we internally add a check that avoid a mod value to bring the main stat below 0, the player Unequipping the item will get a 50 str value instead of his original 40 value.

That's because, unlike spells stat modifier, modStats modifier are not tracked anywhere. 

So the best solution for me is that the scripter is responsible to check if a modStat will bring the stat below 0. (It's a simple script)
```
Example:
//Argn1: ModStr value
[FUNCTION setModStr]
if (<tag0.override.modstr>)
	modstr -= <tag0.override.modstr>
	tag.override.modstr = 
	return 
endif
if ((<str> + <argn1>) < 0)
	argn1 = -<str>
	tag.override.modstr += <argn1>
	sysmessage @50 Modstr overrided to <argn1>  
endif
modstr <argn1>
```

Jhobean:
Seem legit to me. Tag.override.modstr wiil be a custom modstr only in script side?

drk84:
yes

Tolokio:
```
on=@statchange
if <argn1> <= 2 //if str dex or int
    if <argn3> <= 0 //if new value will be negative or 0
        argn3=1 
    endif
endif
```
I think this or something like this is the best way to fix it.
str is <modstr> + <oSTR> so no more would be needed.

cbnolok:
That should be double checked, then if there's still overflow, the wisest thing would be only to limit the stat to 0 before overflowing

Tolokio:
[...]

PR #1258, opened 2024-06-30
MAXHITS FIX

Tolokio:
Well, This is a fix try for maxhits. 

Now u should use @statchange features if u want to manipulate any maxstats. 

Still testing it.

### DIFF
--- a/src/game/chars/CCharStat.cpp
+++ b/src/game/chars/CCharStat.cpp
@@ -377,6 +377,8 @@ void CChar::Stat_SetBase( STAT_TYPE i, ushort uiVal )
 	}
 
 	m_Stat[i].m_base = uiVal;
+    if (i != STAT_FOOD)
+        Stat_SetMax(i, uiVal);
 
 	if ( (i == STAT_STR) && (uiVal < uiStatVal) )
 	{
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3779,3 +3779,6 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 	// ITEM_CANTIMER_IN_CONTAINER	1	// Items can hold a timer even if equipped.
 - Added: item CAN flag (Issue #1180):
 	CAN_I_TIMER_CONTAINED   0x40000000  // This item (if not "sleeping") can have a timer even if inside a container (overrides global default behavior).
+
+30-06-2024, Tolokio
+-Fixed: Maxhits to work normal even if u change it. Fix as well @statchange undesired behaviours.


DavideRei:
In my shard i change maxhits, maxmana ecc. in @create trigger for npcs and directly for players in other scripts, players can have maxhits != str and maxmana != int
Why should we use @statchange ?

Tolokio:
> In my shard i change maxhits, maxmana ecc. in @create trigger for npcs and directly for players in other scripts, players can have maxhits != str and maxmana != int Why should we use @statchange ?

U can still doing it. But maxstat will change if str change. U need to prevent that on statchange. something like this
```
[event e_npcs_statchange]
on=@statchange
if (<argn1>==4)
  if (<maxhits> != <str>)
    Return 1
  endif
endif
```

AtillaG1:
Does that make any changes to default behaviours with using .ini flags or not?

Tolokio:
> Does that make any changes to default behaviours with using .ini flags or not?

give me an example
I tried to do less changes as possible

AtillaG1:
> > Does that make any changes to default behaviours with using .ini flags or not?
> 
> give me an example I tried to do less changes as possible

player STR = MAXHITS 
than you have item properties which can increase any stat (+5 str increase also +5 maxhits, and with unequip it decrease | or +5 maxhits property increase only maxhits) etc...
for chars I have in on@create specified both STR and MAXHITS is higher. But if maxhits or maxstam not defined, it will always be as the STR or DEX etc...

and there is nothing in core that is breaking that or not ignoring the actual value.

I dont really understand this change so if its not really changing this behaviour which was in for long time, just ignore me. 😁

Tolokio:
it changes.

Right now, if u set maxhits it will not change never as @statchange is not going to be fired. That causes problems by other sides.

This change remove that, so if u set maxhits it will be changed again to STR when npc's stats changes.

So, u need to deny maxhits to be changed again using @statchange by script 

If not done, @statchange is not able to deal with maxhits once changed and that gives problems.

AtillaG1:
In this case it would be better to make it as ini flag option or something, since you basically changing whole behaviour and forcing people to make script that will be running on every NPC every X seconds you set your stats to update.

This same discussion was here years ago on the beggining of X and even back than it didnt make sense to force the MAXstats always be as STR or DEX.

I still dont like it. The trigger should not change anything that is set in defs or @create just because. 🤔

Tolokio:
> In this case it would be better to make it as ini flag option or something, since you basically changing whole behaviour and forcing people that use it as i described to make script that will be running on every NPC every X seconds you set your stats to update. I believe it hsould be vice versa, If you need sometime extra, you should be able to make it script wise, but if there is something broken in the trigger, obviously it has to be fixed.
> 
> This same discussion was here years ago on the beggining of X and even back than it didnt make sense to force the MAXstats always be as STR or DEX automatically.
> 
> I still dont like it. The trigger should not change anything that is set in defs or @create just because. 🤔

Right now, @statchange dosnt work as expected. @statchange is very powerful if u use it. Is not fair to let it as it is.

Sphere.ini config option is possible but anyone that want to adapt his live shard to this new feature just need 3 minutes using @statchange. I would prefer to add a "local.oldMaxhits" to the trigger  instead but i dont like it as well.

src/game/chars/CCharStat.cpp
```diff
@@ -295,7 +295,10 @@ uint CChar::Stat_GetSum() const
 ushort CChar::Stat_GetAdjusted( STAT_TYPE i ) const
 {
 	ADDTOCALLSTACK("CChar::Stat_GetAdjusted");
-    return ushort(Stat_GetBase(i) + Stat_GetMod(i));
+    int t = Stat_GetBase(i) + Stat_GetMod(i);
+    if (t < 0)
```
Jhobean, line 299: Should have a warning in console?

Jhobean (commented):
Add comment

src/game/chars/CCharStat.cpp
Tolokio: If there was no warning before I dont see the sense of a warning now that works fine.  But your post makes me think... Why should we care player get zero str due negative modstr?
It can be problematic as maxstr can become 0 and die. You can fix this by @statchange but...

Why dont we set "1" as minimal value for any base stat? Got sense cos 0 str would mean u can even move...
Should we make that str int and dex can't be lesser than 1? For me is better than zero but as can be done by @statchange, idc.

Jhobean: Because if someone have stat ajusted like this, its because some script is bad. Admin should note the negative value to remove it back later.

Tolokio: NOp, dosnt have to be bad scripting. It can be desirable. And in my case it is. Injuris + equipment + debuff + whatever  can make it too often.

Anyway, where is the need to "remove it" if sphere already do it? 

The warning: "hey admin! there is no problem but u should avoid modstr to be negative so i dont have to do it"...
Think about it...

Soulless-1:
wait am i to understand that this change is to make maxhits change once the str changes? if that is the suggestion it is a very bad one. Many creatures do not have their maxhits linked to str. this also works the complete opposite of how Ultima Online is expected. 

right now i do not have any issues with maxhits being changed randomly to the str value. im not sure i understand the problem or the suggestion.

src/game/chars/CCharStat.cpp
Soulless-1: i wouldnt make it so that you force someone to not be able to use 0 for str. people can look out for that in their scripts, no reason for us to make needless checks when they can just be careful.

AtillaG1:
Basically what he is saying from my understanding (I didnt have a time to test myself) is that in trigger @statchange if you want to set maxhits with ARGN3 , it does not go through or it will bug something with STR. 
image
Based on that he is fully changing the behavior of this trigger to automatically correct the maxstats to the stats they are linked to (IF NOT SET) ---- but now it will change even if set.
He is suggesting that it is easy to make script fix for that in trigger to return 1 this behaviour, which if we speak about the optimisations, will fire this code every X seconds on each char that is changing their stats, regening etc... Which can be thousands.

In this topic, there is getting mixed up second problem with negative values being set in the stats or not, I dont think we should mix thing up.

Soulless-1:
well it absolutely should not adjust maxstat to the stat itself if its set to a separate value, that would be very bad. if the maxstat is not set it should always be linked to the stat itself. so to recap how it SHOULD work, and it does at the moment from what i can tell:

if you set maxhits, and you gain STR or set STR it will not adjust the maxhits.
if you have NOT set maxhits and the STR changes, then the maxhits should in fact adjust themselves to the linked str. 

that in my experience is how it already works and has been working fine.

Tolokio:
> well it absolutely should not adjust maxstat to the stat itself if its set to a separate value, that would be very bad. if the maxstat is not set it should always be linked to the stat itself. so to recap how it SHOULD work, and it does at the moment from what i can tell:
> 
> if you set maxhits, and you gain STR or set STR it will not adjust the maxhits. if you have NOT set maxhits and the STR changes, then the maxhits should in fact adjust themselves to the linked str.
> 
> that in my experience is how it already works and has been working fine.

Dude what u are saying is not right but I think I would need to speak a lot show you why... Have been couple days with this and @statchange and maxhits system is just bad coded. And if the code is right, the design is not.
I think u dont realize the problem, try use osi formula.

Osi Formula-> Maxhits = 60+(str/2)

Try it and test it and tell me method u are using.

Maxhits is not supposed to change why? Do u know internal OSI codes? Osi use "maxhits" as an overrride as well? there are many ways to achieve this. If u want to have @statchange capped and bad code just for lazyness im out. Sphere should work with maxhits. 

This fix gives total control of any behaviour on @statchange. Right now statchange is capped. So if u dont want maxstat to change,. u can do it in less than 3 mins. The lazyness of certain admins can't rule our improvements.

Tolokio:
image

see this code and tell me it is right...

Soulless-1:
> > well it absolutely should not adjust maxstat to the stat itself if its set to a separate value, that would be very bad. if the maxstat is not set it should always be linked to the stat itself. so to recap how it SHOULD work, and it does at the moment from what i can tell:
> > if you set maxhits, and you gain STR or set STR it will not adjust the maxhits. if you have NOT set maxhits and the STR changes, then the maxhits should in fact adjust themselves to the linked str.
> > that in my experience is how it already works and has been working fine.
> 
> Dude what u are saying is not right but I think I would need to speak a lot show you why... Have been couple days with this and @statchange and maxhits system is just bad coded. And if the code is right, the design is not. I think u dont realize the problem, try use osi formula.
> 
> Osi Formula-> Maxhits = 60+(str/2)
> 
> Try it and test it and tell me method u are using.
> 
> Maxhits is not supposed to change why? Do u know internal OSI codes? Osi use "maxhits" as an overrride as well? there are many ways to achieve this. If u want to have @statchange capped and bad code just for lazyness im out. Sphere should work with maxhits.
> 
> This fix gives total control of any behaviour on @statchange. Right now statchange is capped. So if u dont want maxstat to change,. u can do it in less than 3 mins. The lazyness of certain admins can't rule our improvements.

The formula is very dependent on the era, and while your formula for OSI you might prefer that doesn't mean that it's correct for everyone, especially people who do not use some of the most modern rulesets, And that should only apply to players. We know for certain that NPCs have their maxhits separated from their str. You can see that by looking at nearly any NPC. Right now the only time that maxhits does not get adjusted is when you set it specifically to be at a certain value, which is VERY much canon with how UO operates. 

The same argument for laziness could easily be applied to your scenario, it could be done in 3 minutes. at very least this should be a toggle, but it seems super needless considering how it's easily 3 lines to do.

on=@statchange
if (<argn1>==0) && !(<src.npc>) 
  src.maxhits=<eval (60+(<argn3>/2))>
endif

Tolokio:
> He is suggesting that it is easy to make script fix for that in trigger to return 1 this behaviour, which if we speak about the optimisations, will fire this code every X seconds on each char that is changing their stats, regening etc... Which can be thousands.

Thousand? why not millions of times per second?

@statchange is a light trigger it dosnt fire too often. Regeneration is not handled by @statchange. Impact on performance is very very low. I dont want to be rude but maybe talking having no idea is not the best things to do. 

It is disgusting to try help and need to be defending my help or explaining everything for people dont even understand the problem but talk as they do. Dudes, I dont need this changes to be done. Whatever u do is fine for me.

During years I Have been talking with some people that use to help but not now, trying to get them back. They will not cos it is very tired to deal with certain people (mostly devs). People use to be disagree always. I dont care if u disagre, but at least TEST IT or be sure to understand the pull before disagree. ty.






> > > well it absolutely should not adjust maxstat to the stat itself if its set to a separate value, that would be very bad. if the maxstat is not set it should always be linked to the stat itself. so to recap how it SHOULD work, and it does at the moment from what i can tell:
> > > if you set maxhits, and you gain STR or set STR it will not adjust the maxhits. if you have NOT set maxhits and the STR changes, then the maxhits should in fact adjust themselves to the linked str.
> > > that in my experience is how it already works and has been working fine.
> > 
> > 
> > Dude what u are saying is not right but I think I would need to speak a lot show you why... Have been couple days with this and @statchange and maxhits system is just bad coded. And if the code is right, the design is not. I think u dont realize the problem, try use osi formula.
> > Osi Formula-> Maxhits = 60+(str/2)
> > Try it and test it and tell me method u are using.
> > Maxhits is not supposed to change why? Do u know internal OSI codes? Osi use "maxhits" as an overrride as well? there are many ways to achieve this. If u want to have @statchange capped and bad code just for lazyness im out. Sphere should work with maxhits.
> > This fix gives total control of any behaviour on @statchange. Right now statchange is capped. So if u dont want maxstat to change,. u can do it in less than 3 mins. The lazyness of certain admins can't rule our improvements.
> 
> The formula is very dependent on the era, and while your formula for OSI you might prefer that doesn't mean that it's correct for everyone, especially people who do not use some of the most modern rulesets, And that should only apply to players. We know for certain that NPCs have their maxhits separated from their str. You can see that by looking at nearly any NPC. Right now the only time that maxhits does not get adjusted is when you set it specifically to be at a certain value, which is VERY much canon with how UO operates.
> 
> The same argument for laziness could easily be applied to your scenario, it could be done in 3 minutes. at very least this should be a toggle, but it seems super needless considering how it's easily 3 lines to do.
> 
> ```
> if (<argn1>==0) && !(<src.npc>) 
>   src.maxhits=<eval (60+(<argn3>/2))>
> endif
> ```
 
Did u test that script deeply? I did some days ago and that script will mess ur server if used.  I already said why u can't use that code somewhere on this pull or discord. IS it fixeable? yes, but gives problems.
MAybe u can apply the lazyness argument to me, but then what should be applied to you? At least I came with homework done. Test that script, look code. Maxhits is bad design right now and mess other things.

Tolokio:
yes, well, Not personal but I have to admit it is tired for me certain situations that makes helping the project disgusting for me and many others. Telling me that Im known for no testing my code is fair, but a personal attack. As I said many times,I TELL I dont test them to force anyone to test them, not only me. Did u find any fail on my "untested code" could be...

Anyway.
U found modstr is not going changing str because of your script. Once maxhits  is set, MODSTR stop working. So, it is fixeable. How would u fix that?

Soulless-1:
Yeah i would just add that into your script and calculate what the str would come to overall with the added argn3 value

```on=@statchange
if !(<src.npc>) 
  if (<argn1>==0)
    src.maxhits=<eval (60+(<argn3>/2))>
  elseif (<argn1>==8)
    src.maxhits=<eval (60+((<src.ostr>+<argn3>)/2))>
  endif
endif

**Comment by Soulless-1** [comment]
modmaxhits just would add to the total maxhits so you dont need to calculate it. feel free to give it a test for your use case.

**Comment by Tolokio** [comment]
> modmaxhits just would add to the total maxhits so you dont need to calculate it. feel free to give it a test for your use case.

I dont understand it, why would it be applied if maxhits is denied?

**Comment by Tolokio** [comment]
IM back:

Well, I´ll share two triggers, both tries to do the same but one is with fixes the other without them.  I did first trigger following your method, I tried to be fair but I dont have much time. plz try to improve or do your own one so both method do the exactly the same things. U will realize the absurd of the complexity.

```





//With New System
on=@statchange
if <tag0.override.maxstat_<argn1>> //This method replace setting maxhits on @create. U can use notepad++ to mass replace
	argn3=<tag0.override.maxstat_<argn1>>
else	
	if (<argn1>==4 && <f_ishumanoid>) //apply osi formula on any humanoid
		argn3=<eval (60+(<argn3>/2))>
	endif
endif
If <argn3>==0 //dont let argn3 be negative or zero. 
	argn3=1
endif


///////////////////VS


//with Old system
on=@statchange
if  <src.maxhits>!=<eval (60+(<str>/2))>// || <tag0.MaxStatIsOverriden_<argn1>>//  How do we know it is being used by npc or player?
	if (<argn1>==0 && <f_ishumanoid>)
		if (<eval (60+((<argn3>+<modstr>)/2))> <= 0) //maybe we get a negative value... good place to deal with it.
			maxhits = 1 // Imagine I want maxhits to never reach "0" or never get under zero, How would you do it? there are diferent ways.
	elseif (<argn1>==8)
		if (<eval (60+((<src.ostr>+<argn3>)/2))> <= 0)
			maxhits = 1 
		else	
			maxhits=<eval (60+((<src.ostr>+<argn3>)/2))>
		endif
	endif
endif

```
Where is the need of that complexity? Replacing "maxhits=" by "tag.override.maxstat_0" is some kind of redflag? to much work? Do we need to have a capped @statchange cos of it?

I dont find right to leave it as it is cos we dont want to change "maxhits" behaviour. USing "maxhits" as an "override of maxhits" was a nice idea to save memory an code in past. Right now it is not cos there is a trigger called @statchange that wants to be unchained.  Right now @statchange is capped and it is a pitty cos @statchange is very well designed and powerful.

**Comment by Soulless-1** [comment]
I do not understand what you are wanting to do. why would you replace maxhits with maxstat_*? what is capped exactly, and what is the end goal? That might help me answer the question better.

**Comment by Tolokio** [comment]
> I do not understand what you are wanting to do. why would you replace maxhits with tag.override.maxstat_*? what is capped exactly, and what is the end goal? That might help me answer the question better.
```
//With New System
on=@statchange
if <tag0.override.maxstat_<argn1>> //This method replace setting maxhits on @create. U can use notepad++ to mass replace
	argn3=<tag0.override.maxstat_<argn1>>
else	
	if (<argn1>==4 && <f_ishumanoid>) //apply osi formula on any humanoid
		argn3=<eval (60+(<argn3>/2))>
	endif
endif
If <argn3>< 0 //dont let argn3 be negative.
	argn3=1
endif
```

Your questions are already answered on this post. I know my english is poor but I think it is not only my english. The goal is to fix @statchange and improve stats system.

tag0.override.maxstat_0 is an override for maxhits. u can replace any stat value using "tag0.override.maxstat_#", IS like "maxhits" but u can use it with any stat. not only maxstats. 

pLZ dont make me writte more. The info is there, the problem is exposed, the fix is done, I answered all question as good I could... If u want to understand just try to do same script I did. I even did an example for u... Just do an osi system as i did..  But a full system with all security checks and that studd, not only a part.

**Comment by Soulless-1** [comment]
This is likely an issue with translation, but i'm looking at the pull request and one of the only unaddressed topics is negative values in stats. is that the only issue here? that seems like a problem, but the solution should not be to set it to 0. we should just allow negative values in stats and skills such as in previous versions of sphere. forcing it to 0 could have unintended affects interacting with other scripts that remove then restore stats/skills.

**Comment by Tolokio** [comment]
> This is likely an issue with translation, but i'm looking at the pull request and one of the only unaddressed topics is negative values in stats. is that the only issue here? that seems like a problem, but the solution should not be to set it to 0. we should just allow negative values in stats and skills such as in previous versions of sphere. forcing it to 0 could have unintended affects interacting with other scripts that remove then restore stats/skills.
> 
> and obviously people should be doing validations prior to editing values, that's just a part of good scripting. but yes i would agree we should not have stats go from 0 to 65535, is that the issue here?

The problem is that sphere transform negative values to ushort.  U can propose whatever u think is the best. The fix dosnt have to be my fix, it is a lazy fix.  I just put the problem and a fix on the table.

So what's ur vote? les negative values being used?
One thing to take in count, the use of ushort and that kind of stuff is not only to limit stats but good memory usage.
If we want negative values. that would need use INT32. Is not a real problem but something to take in mind. Int32 is much bigger.

Int16 -- (-32,768 a +32,767)    //(short)
Int32 -- (-2,147,483,648 a +2,147,483,647)
Int64 -- (-9,223,372,036,854,775,808 a +9,223,372,036,854,775,807)

I think we may let negative values reach @statchange. But stats sholdnt be negative. Negative int or str has no sense. In physics negative str still being str. So -100 str or 100 str would have same power.

if we let stats to be int32, this will be min and max value: -2,147,483,648 a +2,147,483,647

For me, stats shouldnt be negative. Maybe "hits" "mana" or "stam" got sense to have negative values. I could do good stuff with that feature. But str, int or dex no sense.

Whatever the solution is, we need to think right about this, and do best use of memory as possible.

**Comment by cbnolok** [comment]
I find the ushort overflow fix legit and useful. Regarding the trigger firing also for max*stat*, it has always been like that, at least since 56b. It might make sense for someone and it might not for some other people, there's no right or wrong decision, so i don't see any valid reason to change it, it will break compatibility with no benefits since this behavior can be changed easily via scripts.

**Comment by Tolokio** [comment]
> I find the ushort overflow fix legit and useful. Regarding the trigger firing also for max_stat_, it has always been like that, at least since 56b. It might make sense for someone and it might not for some other people, there's no right or wrong decision, so i don't see any valid reason to change it, it will break compatibility with no benefits since this behavior can be changed easily via scripts.


I dont agree with your opinion about maxhits. It is not good to leave it as it is. And the Scriptfix is not that easy. It is easy if u do a simple script without security checks. Try full system plz to understand my point of view as scripter.

Right now, maxhits works as an override of "Stat_GetAdjusted( STAT_TYPE i ) const", is not a real property and that is not good now. Was very smart when it was created cos not only it saves memory but gives more tools to admins when there was no @statchange. Was brilliant but not for actual sphere-x have been past many years and the feature needs to be fixed now.

Sphere 56b scriptpack or any server's script of 56b dosnt work on sphere-x without many fixes. There is no retrocompatibility. And knowing this, I see a mistake to keep certain features like this that caps sphere-x.
We should leave it as real property, and let admins use @statchange to do any override using tags as I do on the next script.

```
//how to override any stat, maxstat,modstr, food if maxhits is fixed.
on=@statchange
if <tag0.statoverride_<argn1>>
      argn3=<tag0.statoverride_<argn1>>
endif
```
I can do the changes on the sphere-x scriptpack to leave it working, it would get me 5 minutes to set this and replace "maxhits=" from npcs.

At least if u are not going to fix |statchange, go spherewiki and update web to warn people that it is not 100% functional, and explain how some weird/unexpected behaviour depending on what u want to do with it and how you do it.


If u want to see diference, This is how a full osi system would look if @statchange is fixed:

```
//full osi for players and npcs with all security checks. (with fixed maxhits)
on=@statchange
if <tag0.override.maxstat_<argn1>> //tag to override any statchange (not only maxstats)
	argn3=<tag0.override.maxstat_<argn1>>
else	
	if (<argn1>==4 && <f_ishumanoid>) //apply osi formula on any humanoid
		argn3=<eval (60+((<ostr>+<modstr>)/2))>
	endif
endif
If <argn3>< 0 //secutiry check, dont let argn3 be negative. 
	argn3=0
endif
```

PLEASE using actual sphere, try to do a functional script thats do the same than the script I did. Until someone do it, I will not feel understood. But do entire system, including the part on equipment @equip @unequip if u count with it. And test it...
Idc if u fix it or not, but makes me feel understood plz.

**Comment by Soulless-1** [comment]
For my edification, can you explain how you view the purpose of maxhits, maxstam, and maxmana as?

**Comment by Jhobean** [comment]
I read all your post Tolokio and not even able to see what the bug is except when they going negative. For all the rest all seem to work like expected. You seem to include different issue and different behavior in your many explanation.

Could you have a Quick and fast bug example instead of long theory?
Limit to:
```
What the bug and quick script to reproduce:
What the behavior it should do:
```

About the files modification you propose, it will create more bug than remove. Actually you will force value to 0 and when player remove the "modifier" stat will be broken.

Tolokio:
ty jhobean

My english is bad, the maxhits case is hard to explain and also to understand and is not related to negative values (looks like it is but not). Right now maxhits affects many things. There is no "one thing" to explain, it is a whole thing.
The change I propose for "maxhits" is a fix for @statchange and an improvement for statsystem. Dont look for a bug.

image

Souless posted this solution for the osi-system, but it is not complete. Where do people check that modstr dosnt makes str going negative? It 

This is the solution drk84 did that many people use:

image

So now it would be "complete".
There are more things to take in mind but why continue? Now im going to put my full script with the fix that includes all things done by soulless + drk scripts and even more.

image

Tolokio:
image

Souless posted this solution for the osi-system, but it is not complete. Where do people check that modstr dosnt makes str going negative? normaly on equipment. I beg him to do full system but he dont. ok ill complete the system for him np. 
ill use the solutions drk84 posted here: https://github.com/Sphereserver/Source-X/issues/963

image

So now it would be "complete". There are more things to take in mind but why continue? It is enought I think.
Now im going to put my full script with the fix that includes all things done by soulless + drk scripts and even more.

image

Tolokio:
Well, if not fixed we should then fix the spherewiki at least.

image

How would it be done? Maxhits read and writte method works diferents. Whatever u put is fine for me.

At @statchange there should be a warning telling why @statchange could not work as expected.

Soulless-1:
After some discussion we've determined that maxSTAT is working as intended. Sorry that this is not aligned with your vision of maxhits, maxstat, etc. 

We're going to go ahead and close this request out. If you would like to open a PR for @statchange negative value, please do so. This pull request has become very muddy.

Tolokio:
well u attacked this modification before understanding it. NOw close it. well it is fine. ty

I repeat, this is not about if maxhits works fine or not. This is about have better tools and code and decide when it is worth or not to remove an old outdated feature.

cbnolok:
We have discussed it thoroughly between us and with you and we all very agree on this. This is nothing personal on anyone's side, simply we don't find it a good idea generally and also explained why. I'm sorry that you don't feel the same way but that's it and honestly it's not the case to fiercely keep fighting on and on on this. Please realize that other people (multiple and experienced in the specific case) might think about stuff differently than you, and that is fine.

Tolokio:
> We have discussed it thoroughly between us and with you and we all very agree on this. This is nothing personal on anyone's side, simply we don't find it a good idea generally and also explained why. I'm sorry that you don't feel the same way but that's it and honestly it's not the case to fiercely keep fighting on and on on this. Please realize that other people (multiple and experienced in the specific case) might think about stuff differently than you, and that is fine.

Agree with you nolok, but who decided and how? ....

 it is ok. if u understand it and you choose knowing what u choose go ahead.

cbnolok:
I decided, after checking the issue myself and asking other people's feedback.

Issue #1248, opened 2024-05-26
can_i_damageable and ItemHitpointsUpdate, do not update status


GladieUO:
Hi,

so i find out that if u have item with can_i_damageable, and you set as GM .xhits 10, it will update the status instantly.

But if you do it via script like hits -=10 it will take hits, but not update the status bar.
If you try use damage <amount>, it will trigger the @damage trig, but not give any dmg.
Is it fixable or client issue?

![image](https://github.com/Sphereserver/Source-X/assets/25097484/3d620f75-0a25-415c-b7bd-df6bc8c98584)

edit: if you change dispid or for example target that item with something, it will update the status, but update doesnt work
edit2: chaging something on the item works too, for example immediately add +1 to more2

Jhobean:
What if you do Update on damage trigger?

AtillaG1:
> What if you do Update on damage trigger?

Seems like its not possible to pass damage to the item with this flag to fire @damage trigger on it.
If I tried to bypass it and then fire update in there it didnt help.

By using damage XY,dmg type,source to trigger @damage it doesnt do damage

Its all weird 😁

PR #1259, opened 2024-07-04
CAN_I_DAMAGEABLE not updating status bar

GladieUO:
Another attempt for the small fix. #1248 also together with this we should consider #1260
The status bar would not get updated when you lower hits on the item with the flag.

**Only way it would get updated is:**
1. reopening status
2. changing some value on item (like MORE1 ++)
3. Targeting item by whatever


Tolokio:
does that fix works?

AtillaG1:
> does that fix works?

does for me. Everytime the hits gets changed from the custom script, statusbar updated. Before it didnt. If its correct fix im not sure.

Tolokio:
Confirmed bug and confirmed that Gladie's fix works. But you lack the fix as well for maxhits
image

I dont know if it is how it should  be fixed but works.

Anyway, I found something weird. maybe is normal I dont know.

image
image

AtillaG1:
I wasnt sure if we need to have it update on maxhits. Maybe if there is like 50/100 hits and you decrease maxhits to 75 so the ratio of status would change too?

Tolokio:
> I wasnt sure if we need to have it update on maxhits. Maybe if there is like 50/100 hits and you decrease maxhits to 75 so the ratio of status would change too? I will update the PR.

You should split commits because there is a fix for follower slots.

Well, I think maxhits is set after the item is created or maybe placed in the world so..., and custom scripts could need it too. For example, a repair system that reduces maxhits but sets hits to 100%.

Anyway, I'm not sure if that's the correct approach or if it's necessary. I think u can force the ticks just giving a timer to the item. let me check

AtillaG1:
> > I wasnt sure if we need to have it update on maxhits. Maybe if there is like 50/100 hits and you decrease maxhits to 75 so the ratio of status would change too? I will update the PR.
> 
> You should split commits because there is a fix for follower slots.
> 
 I Made new branch for this fix but somehow it shows also the followerslots I dunno why 🤐

Tolokio:
U need to do a branch from source-x instead from ur own fork.
image

Then do the changes on that branch and pull. Remember u need to edit changelog as well.

cbnolok:
Did you manage to remove the FOLLOWERSLOTS change from this PR?

AtillaG1:
> Did you manage to remove the FOLLOWERSLOTS change from this PR?

I think I managed to do it right now.

cbnolok:
I'm sorry, could you fix the conflicts please?

GladieUO:
> I'm sorry, could you fix the conflicts please?

should be done

cbnolok:
Manually merged.

PR #1263, opened 2024-07-06
FIx for Negative STAT by negative MODSTAT

Tolokio:
This is the most lazy fix we could do for modstat's issue. (but maybe better than nothing or even other fixes.)

image

**Is there an alternative by Script?**

Yes, Drk84 posted this method for STR and MODSTR. 
image

 You would need one script like this for each modstat and modmaxstat (x6). So the fix is really bigger than this, but that is the method. You would also need to adapt your equip's events to work with the new f_modstat instead the modstat.


cbnolok:
It doesn't compile. In any case, i'll send a different fix for that.

Tolokio:
it was compiling in local. ok

Tolokio:
I dont know if this is really worth, but I thought we can extend @statchange or create a new trigger for this:

image

something lioke this but good done

cbnolok:
In my opinion it's not worth it. That one is already working and doing its job.

PR #1265, opened 2024-07-20
Add dragon breath on @NPCSpecialAction

Jhobean:
Added a new feature about NPC special action.
I tested it well and it work nicely. The bug I had was if you set return 1 when dragon breath, dragon was stuck and always retest to send breath.

I just added a 1 stamina consumption and all work well!


cbnolok:
Could you explain better the need of argo? Also, can we do the stam check via scripts directly instead of hardcoding it? One might want to increase the stam requirement, or use mana or nothing instead.

Jhobean:
Trigger is use 2 place in the source: Argo is needed on dragon breath because you need to know the target.

On the other place (where every wandering NPC ) the trigger appear, no argonis require because there no target. BUT, since I use Argo on my script, I have a invalid argo error each time I use the trigger. The way I found to fix is adding argo to everybone on the trigger.

About the stam, it's not the dragon breath itself consuming stam... Nocspecialaction trigger appear BEFORE the breath. If you set return 1, breath is cancel and next tick the dragon will try again to breath because he his full stam. So the dragon will just freeze for ethernity. The workaround to block this freeze was to consume 1 stam.

cbnolok:
Uhm wait, two questions. First, couldn't you use ARGO.ISVALID to avoid errors?
Then, does the trigger fire at every tick, attack try (like hittry) or what else? I don't get why should we halt the trigger being frequently called by doing STAM -= 1. The npc will get to zero stam by doing nothing, if we do return 1.

Issue #885, opened 2022-08-17
Animations of monsters


Prapilk:
The actions of monsters and animals are governed by mobtypes and mobtype flags. I think Sphere is "stuck" on MONSTER  or ANIMAL fly or not. 

Example: dragons have a 1009 flag which allows them to cast spells with action 12 (stomp in fiddler). But ingame they do action 12 very rarely like normal monsters.

Another example with flags 1000 (c_ophidian_shaman in anim 85, Terathan in 72 etc....), they have the particularity of having action 12 in cast like the dragon BUT ALSO action 13 and 14 in cast. In the game it is not taken into account and it sticks together to hit itself in melee without using actions (again like a simple MONSTER 0 lamda).

mobtypes MONSTER 1000 => action 12, 13 and 14 cast spells.
mobtypes MONSTER 1008 => action fly 19, 20, 21 which are supported but also with action 12 and 13 in caster
mobtypes MONSTER 1009 => action fly 19, 20, 21 which are supported but also with action 12 in caster
mobtypes MONSTER A00  => here the particularity is that there is very little action and that attack actions 4, 5 and 6 replace actions 12, 13 and 14.
mobtypes MONSTER 1004 => actions 7 and 8 are added and action 12 is used as the summoning animation of the monster.

There are other flags but I believe it's client side. The MONSTER 9 is placed on the monsters directed by the body.def (the body.def refers to the monsters with the flags 1009). I think he was supposed to replace the 1009 with less action (maybe remove flying for dungeon dragons) but the flags of the original ID before the replacement in the body.def is not supported count on the clients I use.

Of course the animal flags that are in the MONSTER IDs (0-200) are for those who fly and depending on the flags we add +- actions BUT NOT THAT, I think that the ANIMAL 20 mobtypes transforms the attack 03 (so action 6) in one movement on the water. So the ANIMAL 20 flag should allow the snake and the little frogs to move on the water.

Of course, what I'm saying is purely empirical after having nit-picked a bit in the muls, but I hope I've been explicit enough for us to consider the question.

Jhobean:
We had some good discussion about  this on discord. https://discord.com/channels/354358315373035542/1010257922086928424/1013891598884544544

Apparently sphere never send packet for some anim. Few monster have specific anim and sphere always use stamdars anim.

We should read mobtype flag and depending of npc, send the anim packet differently.

DavideRei:
[...]

PR #1266, opened 2024-07-20
Read mobtypes.txt and handle standard and custom creatures animations

DavideRei:
Read mobtypes.txt (if available among mul files) and cache it to check animations type when using creatures animations.
Now standard anims (SEA_MONSTER, ANIMAL, MONSTER) and UOP anims are handled correctly depending on which actions anim they have on anim files.
About the anims, i checked client latest version and writed down the info here: https://docs.google.com/spreadsheets/d/1kXHctxuHJIuv6yvsA_nzjov_A1IAYkeKpTqx5_c8yM8/edit?usp=sharing

Added alert anim on NPCs when they start combat, pillage anim when they loot something (items on ground or in bodies) and "summon" anim when they are summoned or added. 
I'm still dubious about the SetTimeout command after summon action. 
I added it to show the entire animation, because if the creature moves, the animation is interrupted. 
Some monsters have 2-3 sec. animation but the timeout freezes them until is completed, so for example if a player summon a creature he can't command it until the timeout is completed. 
What are your thoughts about this? Is there a bettere way to do it?

Now custom anims can be handled by scripts using ANIM command (It was almost useless before, so i decided to replace it) under CHARDEF, to set which anim actions they have/don't have, using the following flags.

MONSTER (H slot) default configuration is with actions: 0 (Walk), 1 (Stand), 2 (Die1), 3 (Die2), 4 (Attack1), 5 (Attack2), 6 (Attack3), 10 (GetHit), 11 (Pillage), 15 (Block Right), 16 (Block Left), 17 (Idle), 18 (Fidget)
    No DIE2 action		= 01,
    No ATTACK1 action   = 02,
    No ATTACK2 action   = 04,
    No ATTACK3 action   = 08,
    No GETHIT action    = 010,
    No PILLAGE action   = 020,
    STOMP action		= 040,
    CAST2 action        = 080,
    CAST3 action        = 0100,
    No BLOCKR action    = 0200,
    No BLOCKL action    = 0400,
    No FIDGET1 action   = 0800,
    No FIDGET2 action   = 01000,
    FLY action          = 02000,
    Anims with Attacks 1-2-3 in actions 9,10,11 and with GetHit in action 20 (there are many custom WOW anims with this configuration) = 0x10000

ANIMAL (L slot) default configuration is with all actions: 1 - 12
    No EAT action       = 01,
    No ALERT action     = 02,
    No ATTACK2 action   = 04,
    No GETHIT action    = 08,
    No FIDGET1 action   = 010,
    No FIDGET2 action   = 020,
    No LIEDOWN action   = 040,
    No DIE2 action      = 080

So for example if you have a MONSTER (H slot) flying custom anim, with Cast2 action but not Pillage action you can set ANIM=020A0
If you have a custom anim with default configuration, you don't need to set ANIM= (i checked with my custom anims to use the most suitable config to don't have to change all chardefs)

Issue: #885


Jhobean:
Good job!  Just want to link the related issue: https://github.com/Sphereserver/Source-X/issues/885

src/game/uo_files/CUOMobtypes.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    CUOMobTypesType mobTypesRow = {0,0};
+
+    _mobTypesEntries.clear();
+
+    CSFileText csvMobTypes;
+
+    if (g_Install.OpenFile(csvMobTypes, "mobtypes.txt", (word)(OF_READ | OF_TEXT | OF_DEFAULTMODE)))
+    {
+        _mobTypesEntries.resize(4096);
```
cbnolok, line 26: Is 4096 an arbitrary number or some hardcoded client limit?

src/game/uo_files/CUOMobtypes.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+                int len = (int)tmpString.length();
+                len = Str_TrimEndWhitespace(tmpString.data(), len);
+
+                if (len == 0 || tmpString[0] == '#') //Empty line or commented
+                    continue;
+
+                //Split the string
+                char* pch;
+                pch = strtok(tmpString.data(), " \t");
```
cbnolok, line 55: You could check, if you like so, Str_Parse. It's a Sphere function which hasn't the drawbacks of strtok, mainly it's not thread safe.

src/game/uo_files/CUOMobtypes.h
```diff
@@ -0,0 +1,62 @@
...
+#define PACK_NEEDED
+#else
+    // GCC based compiler you can add:
+#define PACK_NEEDED __attribute__ ((packed))
+#endif
+
+/**
+* mobtypes.txt
+*/
+struct CUOMobTypesType
```
cbnolok, line 25: You could simplify this as we don't need structure packing here. It's usually needed when we want to read a sequence of bytes (from a UO mul file) and with those fill an entire struct. Structure packing instructs the compiler to not add padding bytes.
Pro read for more detail: http://www.catb.org/esr/structure-packing/
Faster answer: https://stackoverflow.com/questions/4306186/structure-padding-and-packing

cbnolok (commented):
That's an excellent one, thanks!
I'd suggest some little changes before merging it. One of them would be to add a ini setting to enable/disable mobtypes loading, so giving the choice to use the old behavior

src/game/uo_files/CUOMobtypes.cpp
DavideRei: Like this?
```
//Split the string
std::vector<tchar*> splitArray;
splitArray.resize(3);
size_t iQty = Str_ParseCmds(tmpString.data(), splitArray.data(), 3, " \t#");
```

DavideRei:
> That's an excellent one, thanks! I'd suggest some little changes before merging it. One of them would be to add a ini setting to enable/disable mobtypes loading, so giving the choice to use the old behavior

About the ini setting, I thought it wasn't required because the mobtypes file is not loaded (and in this case Sphere uses the old behavior to distinguish between monster and animal) unless it's in the mul folder.
Therefore if someone doesn't want to load it, he could simply remove it. Usually if you are running Sphere on a server you don't need the full installation but only few files: hues, map, multi, radarcol, statics and tiledata.
In addition, if you have the mobtypes file why not use it?

src/game/uo_files/CUOMobtypes.cpp
cbnolok: Yes, or you could just use tchar* ppColumns[3] in place of that vector

src/game/uo_files/CUOMobtypes.h
cbnolok: Exactly

cbnolok:
One could simply just want to use their full client folder also for sphere. In any case, i feel that an ini setting is more straightforward and simple to manage for people wishing to keep the old "buggy" behavior just because it's "classic sphere".

DavideRei:
Done but how can I resolve merge conflicts now?

cbnolok:
Manually merged, needed in order to fix conflicts.

PR #1276, opened 2024-08-22
Add "getting started" documentation

Otacon:


Jhobean:
Very complete. Awsome!

I would not put the sphere version link there at all. Since it's in the X repository, dont need to give option for the old 56d branch.

Otacon:
> Very complete. Awsome!
> 
> I would not put the sphere version link there at all. Since it's in the X repository, dont need to give option for the old 56d branch.

Yeah! I've thought about that as well with the same reasoning behind it :)
I'll remove it. :P

README.md
```diff
@@ -12,6 +12,9 @@ Ultima Online game server, developed in C++.
 | :---: |
 | [![Discord Shield](https://discordapp.com/api/guilds/354358315373035542/widget.png?style=shield)](https://discord.gg/ZrMTXrs) |
 
+## Getting Started tutorial
+If you're new to the Sphere server and want to setup your first shard, <a href="https://github.com/Sphereserver/Source-X/blob/master/docs/Getting-started.md">this is your go-to guide</a>!
```
raydienull, line 16: Change to use local reference, instead of url

`[this is your go-to guide](docs/Getting-started.md)`

Otacon: It didn't work because I was using the wrong syntax :man_facepalming: . Fixed.

Issue #1303, opened 2024-10-10
Packet 0xBE - assistversion [Feature Request]


GladieUO:
I think it would be very beneficial to have a packet 0xBE https://docs.polserver.com/packets/index.php?Packet=0xBE
However Im not sure how the packet gets to the assistant and not the client.
Im trying to check assist version with ClassicUO and Razor, Razor will send correct data, but it seems like the client itself cant handle that and eventually crashes.

Also other clients and assistants might work differently.

Any thoughts?

PR #1304, opened 2024-10-10
Registration of Packet 0xBE

GladieUO:
Just definition or registration of packet to avoid console error of unknown packet. Not sure if its that easy.
I tested it and it dont show unknown packet error anymore. Is that something we can add?
#1303 
https://docs.polserver.com/packets/index.php?Packet=0xBE


cbnolok:
Could you fix that warning?
It means that you need to define the constructor in the cpp file, even by doing PacketAssistVersion::PacketAssistVersion : Packet(......................) = default;

GladieUO:
> Could you fix that warning? It means that you need to define the constructor in the cpp file, even by doing PacketAssistVersion::PacketAssistVersion : Packet(......................) = default;

should be good now.

PR #1305, opened 2024-10-11
Add @DisplayName to be able control the name show on status and Paperdoll

Jhobean:
I added a new trigger for RP server.

With this trigger you are able to control the name you want to display on the paperdoll and the status bar.


cbnolok:
Isn't TAG.NAME.ALT working for you?

Jhobean:
> Isn't TAG.NAME.ALT working for you?

No because, the person knowing you must see the real name. I want hide my name only to the person not knowing me. (Like many RP server do on servUo do)
With the trigger, I can check my tag system in script and decide if this player have condition to see my name

cbnolok:
I'd opt for a different solution. It would be wise to not call any triggers right in the packet class. I'd add that code instead in  CChar::GetName method. Also, is RETURN 1 really necessary? I mean, we could read anyways ARGS and use that

Jhobean:
The problem in get name method is  I dont know who I get the name for.

I mean, my name will be different if I dont know you. If I use getName method I must add argument to the method (pTarget)...

Adding an argument is acceptable?

Jhobean:
Aboit the return 1 it was require because If I dont return 1, We get all the namenwith the title.

Hector the magician
<Name> <title>

If we return 1 we force to shiw only the name  (similar to incognito)

Jhobean:
Just notice I must add the trigger on the Tooltip too.... I'll convert to draft for now.
I just dont figure how to modify GetName function to be able do what I do.

cbnolok:
You can create a CChar::GetNameFor method

Jhobean:
> You can create a CChar::GetNameFor method

I will need to determine when use it (Will requiere a trigger) and I will need to block the rest of Info on paperdoll (Title etc.) (Will need trigger or a bool)

cbnolok:
While it could be useful, i wouldn't keep it in this state for some reasons: i'd avoid triggers in packet code and the removal of the tooltip cache. Having to regenerate it for each char every time can be time consuming for the cpu when the area is very  crowded.

Jhobean:
About code in packet, I habe not other idea for doing it. We should bring all the code of packet outside of the packet method....  It would be the same and make no sense.  Why it's mad to add code there? 

About the tootip, it's just impossible to keep player name in tootip if we cache it.
An iptimisation I can do, is dont cache the tooltip only if you use the displayname trigger.  But since we only dont cache the Player tooltip, I think it's not really game changer

PR #1310, opened 2024-10-16
Added spherecustom.ini definition

xwerswoodx:
- Added: spherecustom.ini definition has been added to Sphere.

1. This allows you to override Sphere settings without touching the Sphere.ini file.

2. For example, when you want to switch from an older Sphere version to 56x or when you haven't updated for a long time, instead of checking all the new data added to Sphere.ini one by one, you can simply use your old sphere.ini as spherecustom.ini. This way, newly added data will take the default settings from Sphere.ini, while your previous custom settings will override the defaults without requiring any further adjustments.

3. On the other hand, it helps you maintain a clean Sphere.ini. Instead of directly changing the Sphere.ini settings, you can create a spherecustom.ini file and add only the settings you want to change there, allowing you to easily track the changes without modifying Sphere.ini.

4. Here's an example of a spherecustom.ini:
```
[SPHERE]
AGREE=1

// Server Data
ServName=MyCuteServer
ServIP=255.255.255.255
ServPort=2593

// Server Settings
AdvancedLos=2
GuardsInstatntKill=0
FeatureT2A = 01|02

[SERVERS]
MyCuteServer
255.255.255.255
2593

[EOF]
```

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3808,3 +3808,136 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 	New values:
 	SPELLFLAG_FIELD_RANDOMDECAY     02000
 	SPELLFLAG_NO_ELEMENTALENGINE    04000
+
+10-08-2024, DavideRei
+- Added: read mobtypes.txt (if available among mul files) and cache it to check animations type when using creatures animations
+- Added: alert anim on NPCs when they start combat, pillage anim when they loot something (items on ground or in bodies) and "summon" anim when they are summoned or added
+- Modified: now standard anims (SEA_MONSTER, ANIMAL, MONSTER) and UOP anims are handled correctly depending on which actions anim they have on anim files (checked on last client version).
+	Please note that some actions are handled by the client (for example walk, run, fly, getHit while flying, die) and they could not work depending on which client you are using.
+- Added: missing/new actions to be used with .ANIM command.
+    ANIM_ALERT              = 35,
+    ANIM_THROW              = 36,
+    ANIM_PILLAGE            = 37,
+    ANIM_STOMP              = 38,
+    ANIM_FLY                = 39,
+    ANIM_LAND               = 40,
+    ANIM_GETHIT_AIR         = 41,
+    ANIM_SUMMON             = 42,
+    ANIM_SPECIAL            = 43,   //UOP anims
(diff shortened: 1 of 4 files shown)


Jhobean:
That mean if value are at 2 places server will priorize spherecustom.ini?

xwerswoodx:
Yes, if you put any data in spherecustom.ini, will override the value in sphere.ini, for example if you put ServName to MyCuteServer like in example, even it's MyShard in sphere.ini, <SERV.NAME> will return as MyCuteServer

xwerswoodx:
I don't know why linux x86 failed to build as I just cleaned duplicated line, probably auto-merge issue so I leave it like that.

xwerswoodx:
> Member

I added that because I thought same but while I was testing Sphere crashes automatically without giving an error on launch when you just put 

```[SPHERE]
AGREE=1

// Server Data
ServName=MyCuteServer
ServIP=127.0.0.1
ServPort=2593

ScpFiles=scripts/
WorldSave=save/
AcctFiles=accounts/
MulFiles=C:/Program Files (x86)/Electronic Arts/Ultima Online Classic/
Log=logs/

[SERVERS]
MyCuteServer
127.0.0.1
2593

[EOF]```

values inside.

**Comment by xwerswoodx** [comment]
Okay anyway I noticed the issue is because of one/couple of these definations;

```
Map0=7168,4096,-1,0,0		//ML size
Map1=7168,4096,-1,1,1		//ML size
Map2=2304,1600,-1,2,2
Map3=2560,2048,-1,3,3
Map4=1448,1448,-1,4,4
Map5=1280,4096,-1,5,5
UseMapDiffs=0
MapViewRadar=31
MapViewSize=18
MapViewSizeMax=24
```

Because it launched when I added these values to sphere.ini, I will cancel the spherecustom.ini and take a look why one of these causes a crash later on.

File: src/game/items/CItemBase.cpp [hunk 1]
```diff
@@ -1032,7 +1032,7 @@ int CItemBase::GetMakeValue( int iQualityLevel )
 
 	CValueRangeDef values(m_values);
 
-	if ( m_values.m_iLo == INT64_MIN || m_values.m_iHi == INT64_MIN )
+	if ( m_values.m_iLo == INT32_MIN || m_values.m_iHi == INT32_MIN )
```
Jhobean, line 1035: This line make my server crash on boot when i'm in debug or on Nightly

image

Issue #1313, opened 2024-10-18
t_multi_addon and redeed doesnt work


GladieUO:
With attempt to redeed multi with t_multi_addon, all components are deleted, but the main item stays untouched and its not redeeded.

PR #1314, opened 2024-10-19
t_multi_addon redeed fix

GladieUO:
Issue #1313
The code was expecting the pOwner to be defined, but since the multi with t_multi_addon are mostly used for the house decoration, placed by players independently from deeds, we also need to check for the player doing the redeeding, since there is no owner on this multi.

The lack of owner would cause the code to be returned 

as always, any revision of my fix is appreciated, as it might not be perfect.


cbnolok:
Could you add a changelog entry, and change uidChar to a name more meaningful, like uidReededingChar?

GladieUO:
> Could you add a changelog entry, and change uidChar to a name more meaningful, like uidReededingChar?

should be good to go. Dont know why is changelog in conflict, I added all the old entries.

GladieUO:
> > Could you add a changelog entry, and change uidChar to a name more meaningful, like uidReededingChar?
> 
> should be good to go. Dont know why is changelog in conflict, I added all the old entries.

Ok not yet, something is wrong 😁

cbnolok:
Something changed in the build system, i'll try to fix that

PR #1319, opened 2024-10-25
New optionflag OF_PetBehaviorOwnerNeutral

canerksk:
If this setting is enabled, pets you own will appear to you with the original noto (old behavior). If this setting is not enabled, the assets you own will always appear natural to you. This setting only applies when the owning character is looking at the asset they own. It does not apply when someone else is looking at the asset you own.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3914,3 +3914,7 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 
 13-10-2024, Jhobean
 - Added: @PetRelease trigger work like @petdesert and return 1 to prevent the pet from being released.
+
+25-10-2024, canerksk
+- Added: New optionflag OF_PetBehaviorOwnerNeutral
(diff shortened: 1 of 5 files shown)


cbnolok:
It looks like there are other undocumented changes in this commit, relative to npc movement

canerksk:
> It looks like there are other undocumented changes in this commit, relative to npc movement

Yes, all changes are included in changlogs.txt.

GladieUO:
> > It looks like there are other undocumented changes in this commit, relative to npc movement
> 
> Yes, all changes are included in changlogs.txt.

so what does the new flag do? Its not mentioned in changelog

Jhobean:
What do the new style walk?

canerksk:
I didn't give too much detail assuming that everyone knows the old style. After MOVEDELAY came, all movement calculations and SetTimeOut method changed. I added the old version before MOVEDELAY came as an option.

cbnolok:
Could you add in the changelog a brief but clear explanation of the difference between the old and the "new" movement style? It would be great also if you could specify which previous sphere version is the term "old" related to, like 56b, 51a, etc

cbnolok:
Could you add the required info please, so that this PR could be merged?

cbnolok:
Merged with commit https://github.com/Sphereserver/Source-X/commit/fc6c6cde151834c82f39565a40653abd05ed24e2

PR #1320, opened 2024-10-25
Added more server function

canerksk:
f_onserver_mode

f_onserver_resync_start
f_onserver_resync_restart
f_onserver_resync_failed
f_onserver_resync_success
f_onserver_resync_finish

f_onserver_save_stage
f_onserver_save_force
f_onserver_save_try

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3914,3 +3914,73 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 
 13-10-2024, Jhobean
 - Added: @PetRelease trigger work like @petdesert and return 1 to prevent the pet from being released.
+
+25-10-2024, canerksk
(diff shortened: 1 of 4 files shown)


cbnolok:
- Could you explain better when is f_onserver_resync_restart called? Only when resuming from a resync pause?
- Isn't f_onserver_resync_finish redundant, since we have success and failed?
- I'd also explain more precisely to users the difference between f_onserver_save_stage, force and try.
- I'd rename f_onserver_mode to f_onserver_mode_change, if you don't mind.
- It looks also that there's f_onserver_exit_later, not documented in the changelog. Why should that be needed when we already have f_onserver_exit? Only for having the exitflag? Be wary that doing scripted stuff at that stage will make the server crash in a ton of cases.

PR #1321, opened 2024-10-25
Added more server (maria db) function

canerksk:
f_onserver_mdb_connect
f_onserver_mdb_query
f_onserver_mdb_execute
f_onserver_mdb_close
f_onserver_mdb_tick


cbnolok:
There might be a slight misunderstanding here, the mdb object is for the in-memory sqlite database, ldb for the classic sqlite file database, while for mysql/mariadb we use the db object

PR #1322, opened 2024-10-25
Logout trigger for NPCs

canerksk:
@Logout trigger When NPCs (mounts) are logged out of the session in any way, i.e. mounted or minimised, they will log out (disconnect) but will not trigger.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3914,3 +3914,6 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 
 13-10-2024, Jhobean
 - Added: @PetRelease trigger work like @petdesert and return 1 to prevent the pet from being released.
+
+25-10-2024, canerksk 
+- Added: @Logout trigger When NPCs (mounts) are logged out of the session in any way, i.e. mounted or minimised, they will log out (disconnect) but will not trigger.
--- a/src/game/chars/CChar.cpp
+++ b/src/game/chars/CChar.cpp
@@ -528,6 +528,14 @@ void CChar::SetDisconnected(CSector* pNewSector)
         GetClientActive()->GetNetState()->markReadClosed();
     }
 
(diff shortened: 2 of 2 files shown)


Jhobean:
Since it's a particular situation on a particular NPC, should we really use the @logout trigger?

I mean, we disconnect the mount when we climb on it to avoid it take damage and maybe more reason. If the purpose is really explicitly for this situation, could we use something like @Dismounted and place it directly on the dismount method? Juste after the @dismont for player.

canerksk:
> Belirli bir NPC'deki belirli bir durum söz konusu olduğundan, @logout tetikleyicisini gerçekten kullanmalı mıyız?
> 
> Yani, üzerine tırmandığımızda hasar görmesini önlemek için montajı ayırıyoruz ve belki de daha fazla sebep. Eğer amaç gerçekten bu durum için açıkça belirtilmişse, şöyle bir şey kullanabilir miyiz?@Dismountedve onu doğrudan dismount yöntemine yerleştirin? Juste'den hemen sonra@Dismontoyuncu için.

In fact, a mount can be logged out not only when it is mounted but also when we directly .xdisconnect it to the mount. Previously, this trigger was not triggered when a mount was disconnected with .xdisconnect.

canerksk:
> Since it's a particular situation on a particular NPC, should we really use the @logout trigger?
> 
> I mean, we disconnect the mount when we climb on it to avoid it take damage and maybe more reason. If the purpose is really explicitly for this situation, could we use something like @Dismounted and place it directly on the dismount method? Juste after the @Dismont for player.

In fact, a mount can be logged out not only when it is mounted but also when we directly .xdisconnect it to the mount. Previously, this trigger was not triggered when a mount was disconnected with .xdisconnect. 

It also disconnects when the mount is shrink.

cbnolok:
That's interesting. Do you have any use cases in mind?

PR #1324, opened 2024-10-26
New @Deposit/@ItemDeposit trigger

canerksk:
SRC/I = Char (Player)
	ARGO = Gold
	RETURN 1 = Prevents depositing gold into the bank.
	It works just like DROPON_ITEM/ITEMDROPON_ITEM.
	When using the virtual gold feature, there was no trigger triggered when physical gold was deposited into the bank, the Dropon_Item was not working because it was triggered lower down, and instead of moving the Dropon_Item up, a separate trigger was added to track the money on the server more easily from another location.

Tested.


cbnolok:
If this should be only for virtual gold going into our bank account, shouldn't it be called something like VirtualGoldDeposit ro VirtualGoldDepositBank? What do you think?

canerksk:
> If this should be only for virtual gold going into our bank account, shouldn't it be called something like VirtualGoldDeposit ro VirtualGoldDepositBank? What do you think?

If virtualization is made for copper or silver coins in the future, it can be used for all three types of currencies.

Jhobean:
What I understant is it working for all gold. Virtual or not.

cbnolok:
> > If this should be only for virtual gold going into our bank account, shouldn't it be called something like VirtualGoldDeposit ro VirtualGoldDepositBank? What do you think?
> 
> If virtualization is made for copper or silver coins in the future, it can be used for all three types of currencies.

You are right, maybe then something like VirtualCurrencyDeposit? The "Deposit" part is fine, but alone i fear it's a bit misleading, one could think it's related to any item.

PR #1333, opened 2024-11-11
@Eat trigger stat increases are not disabled

canerksk:
Trigger was triggered after values ​​were changed and hits, stam, mana gain could not be turned off.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3914,3 +3914,6 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 
 13-10-2024, Jhobean
 - Added: @PetRelease trigger work like @petdesert and return 1 to prevent the pet from being released.
+
+-11-11-2024, canerksk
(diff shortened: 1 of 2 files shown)


cbnolok:
Could you please explain better why does this need to trigger twice?

cbnolok:
Also we need to avoid integer overflows, i'll provide a saturating add function (i think i only wrote a saturating subtraction one). Anyways, i might be wrong, but isn't it there a method that changes the given stat value and does that overflow check itself?

canerksk:
Dolore soluta a atqu

> Lütfen bunun neden iki kez tetiklenmesi gerektiğini daha iyi açıklayabilir misiniz?



> Could you please explain better why does this need to trigger twice?

In order to get the food values ​​before and after eating, it is unclear how much food value there is after eating under this trigger since it is triggered before eating in its initial state.

canerksk:
> Could you please explain better why does this need to trigger twice?

In order to get the food values ​​before and after eating, it is unclear how much food value there is after eating under this trigger since it is triggered before eating in its initial state.

cbnolok:
I'd be cautious with changing how a trigger is called, since it can have profound compatibility issues with old scripts. I'm more open to add an @AfterEat trigger, even if i'd ask why couldn't i get the new value with something like local.newhits = <max <maxhits>, <hits>+<local.hits>>

PR #1342, opened 2024-12-04
Added new CombatFlag ini setting COMBAT_ATTACK_NOAGGREIVED

canerksk:
In the old behavior (55i/55r/56b/56c) we were not guilty when we hit a murderer. Now when we attack a criminal or murderer we are guilty towards that person. An ini setting option has been added for this behavior. Although this is a normal behavior in OSI, it can be perceived as a problem since not every server designs games like OSI.


Jhobean:
Interesting add on!

canerksk:
> Interesting add on!

It wasn't interesting for 20 years, is it interesting now? 😂

canerksk:
> I personnaly dont play with noto so for me it's useless but its a good feature and goos example of .ini utility.

These things seem small but are important to make the emulator usable and appealing to everyone. SphereServer is an emulator that has been around for a very long time and over that time, SphereServer users have some habits. We don't give up these habits easily. As a result, it is SphereServer that brings these habits to the community.

PR #1363, opened 2024-12-29
Reset Item Link when set empty

DavideRei:
In this way you don't have to set Link=04fffffff when you want to remove a link from an item.
You just do Link=


Jhobean:
I like it!!

PR #1364, opened 2024-12-29
Update sphere.ini

canerksk:


Jhobean:
AGREE defauilt value should be 0?

PR #1368, opened 2025-01-13
common/CScriptObj: use posix_spawn() instead of vfork()/execlp()

MaxKellermann:
Using vfork() directly is extremely fragile, and posix_spawn() is easier to use.  This implements a TODO comment.

### DIFF
--- a/src/common/CScriptObj.cpp
+++ b/src/common/CScriptObj.cpp
@@ -28,9 +28,14 @@
 #else
 #   include <sys/wait.h>
 #   include <errno.h>	// errno
-//  #include <spawn.h>
+#   include <spawn.h> // for posix_spawn()
 #endif
 
+#ifdef __APPLE__
+/* this declaration does not exist on macOS, but we need it for the
+   posix_spawn() call */
+extern char **environ;
+#endif
 
 class CStoneMember;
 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
No, I admit I havn't tested it because I don't know how to call it. I changed this only because Sphere wouldn't compile with `vfork()`:
```
FAILED: CMakeFiles/spheresvr_release.dir/src/common/CScriptObj.cpp.o 
[...]
In file included from Debug/CMakeFiles/spheresvr_release.dir/cmake_pch.hxx:9,
                 from <command-line>:
src/common/CScriptObj.cpp: In member function ‘virtual bool CScriptObj::r_WriteVal(lpctstr, CSString&, CTextConsole*, bool, bool)’:
src/common/CScriptObj.cpp:427:9: error: variable ‘inLocalBlock’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
  427 |         EXC_TRY("WriteVal-Ref");
      |         ^~~~~~~
src/common/CScriptObj.cpp:427:9: error: variable ‘inLocalBlockCnt’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
  427 |         EXC_TRY("WriteVal-Ref");
      |         ^~~~~~~
cc1plus: all warnings being treated as errors
```
`vfork()` is indeed a minefield and nearly impossible to use safely without resorting to assembly, because you never know what the C++ compiler generates. But what Sphere does here is illegal, not even counting for compiler quirks. I wonder how this could ever have worked at all.
So something had to be done.

I thought the function was buggy because it left zombie processes, but then I saw the `SIGCHLD` handler which calls `waitpid()` from within the signal handler. I didn't know this was legal (maybe it isn't). But this can obviously break the `waitpid()` call in `CScriptObj` (maybe not if Sphere takes care to adjust the signal block mask before calling `CScriptObj`, but I guess it does not).
So the function is indeed buggy after all, but differently buggy than I thought.

cbnolok:
Which compiler are you using? It compiles fine here with clang and gcc, but i get the point and you might be using a newer version than mine.
Anyways, from here (https://www.gnu.org/software/gnulib/manual/html_node/environ.html) and here (https://man7.org/linux/man-pages/man7/environ.7.html) i think that's not fully portable, a solution might be using something similar, if you agree:
```
#if defined(__APPLE__)
    #include <crt_externs.h>
    #define ENVIRON (*_NSGetEnviron())
#elif defined(_GNU_SOURCE)
    #include <unistd.h>
    extern char **environ;
    #define ENVIRON environ
#else // *BSD?
    extern char **environ;
    #define ENVIRON environ
#endif
```

MaxKellermann:
Oh no, macOS is no longer POSIX-compliant?

cbnolok:
I think it still is, but confusingly enough the POSIX standard seems to just require it `to be declared by the user if it is to be used directly` (https://pubs.opengroup.org/onlinepubs/9799919799/functions/environ.html), how you get it is kinda a convention which can change between OSes (and even OS versions).
Linux mandoc says: `Historically and by standard, environ must be declared in the user program.  However, as a (nonstandard) programmer convenience, environ is declared in the header file <unistd.h> if the _GNU_SOURCE feature test macro is defined.
GNUlib(c?) goes a bit deeper for MacOS, less for *BSD (https://www.gnu.org/software/gnulib/manual/html_node/environ.html). It looks that `environ` might still be used on Mac, but might get slightly more painful in some corner cases (not ours, though, since we aren't using this code in a shared library) and maybe it change again with newer OS versions

MaxKellermann:
Ah "in shared libraries" - then it's still a POSIX violation but not something Sphere needs to care about. I believe it's not necessary to have this big `#if` block, but if you want it, I'll add it. My suggestion is to change my `#ifdef __APPLE__` to `#ifndef __GLIBC__` so all BSDs are covered, as this variable already declared is a GNUism.
I wouldn't check for `_GNU_SOURCE` - it's a macro to be defined by the application, not by the library. All applications that want to use GNU features should set it, and they can even set it without knowing whether they're really on glibc, therefore the presence of the macro isn't a good check.

cbnolok:
I agree then, i'll gladly merge the commit with your newly proposed change

MaxKellermann:
> I agree then, i'll gladly merge the commit with your newly proposed change

Amended (and rebased).

PR #1372, opened 2025-01-20
Add a warning line when an item is delete on save

Jhobean:
On some situation, I had a i_memory deleted on save because there was no color on the i_mem!!!

No warning and nothing!  After 2 hour of debug, I found what was happen and it was missing a log error.

With this line of code the error now appear on log.
image

### DIFF
--- a/src/game/CWorld.cpp
+++ b/src/game/CWorld.cpp
@@ -502,6 +502,7 @@ int CWorldThread::FixObj( CObjBase * pObj, dword dwUID )
 			CItem * pItem = dynamic_cast <CItem*>(pObj);
 			if ( pItem != nullptr && pItem->IsType(IT_EQ_MEMORY_OBJ) )
 			{
+                ReportGarbageCollection(pObj, iResultCode);
 				pObj->Delete();
 				return iResultCode;
 			}


cbnolok:
Can you create a pr for the dev branch instead?

Jhobean:
On my opinion, this is a fix wiith no impact and admin shoukd benefit of thia fix soon as possible. No risk to broke something.
Why shoukd I merge on dev and admin willl use thia fix in many weeks?

cbnolok:
Because i'd like to have that warning on both branches :) 



Il giorno 21 gen 2025, 12:46, alle ore 12:46, Jhobean ***@***.***> ha scritto:

Jhobean:
Master is not suppose to be merge on Dev occasionaly. I think it was what @raydienull was saying.

cbnolok:
Yes the intent was to push single commits and not the whole dev branch to master. Anyways, if @raydienull agrees, we could merge it to master, then merge a separate pr to dev

raydienull:
in this moment, there is no problem in uploading small changes in master. As they are possible improvements/fixes to the stable version.

PR #1375, opened 2025-02-03
CWorldTicker and other minor fixes.

cbnolok:
This *should* fix the missing status updates issue.


cbnolok:
Before merging it, i'd like to have different people to try this and have extensive tests.
If this works as intended/expected, before merging i'll remove some asserts and checks, in order to remove that overhead.

Jhobean:
Seem stable

PR #1379, opened 2025-02-14
.github/workflows/build*: pass "-v" to ninja

MaxKellermann:
The CI log should be verbose so we can see what commands are really being used, and it's pointless to use the default "beautified" ninja output.  In this case, I'd like to know why my ccache changes were not used ("Hits: 0 / Uncacheable: 210"), see commit b06a46ebba1edfef

### DIFF
--- a/.github/workflows/build_linux_x86.yml
+++ b/.github/workflows/build_linux_x86.yml
@@ -102,7 +102,7 @@ jobs:
               -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
               -S ./ -B ./build
     - name: Ninja
-      run: ninja -C build
+      run: ninja -C build -v
     - name: Create package
       run:  |
             mkdir accounts logs save scripts
--- a/.github/workflows/build_linux_x86.yml
+++ b/.github/workflows/build_linux_x86.yml
@@ -103,6 +103,8 @@ jobs:
             cmake -G "Ninja" -DCMAKE_BUILD_TYPE="Nightly" \
               -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains/Linux-GNU-x86.cmake \
               -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+              -DCMAKE_C_FLAGS="-fpch-preprocess" \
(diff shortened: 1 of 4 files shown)


MaxKellermann:
With "sloppiness", all "Uncacheable" are gone.

cbnolok:
I doubt that it will ever be relevant, but what about avoiding to override default CMAKE_C_FLAGS and CMAKE_CXX_FLAGS? Time ago i added some custom variables to inject the flags via cli, with the last commit i renamed them to CUSTOM_C_FLAGS, CUSTOM_CXX_FLAGS, CUSTOM_EXE_LINK_FLAGS

MaxKellermann:
I don't understand. Why would you want to discard the values of user-specified cmake flags variables and instead define new ones with different names?

cbnolok:
As far as i know, if you manually define CMAKE_<LANG>_FLAGS, CMake won't add its default flags

MaxKellermann:
What default flags do you mean?

MaxKellermann:
The wording in https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html doesn't sound like this omits any other flags.

cbnolok:
I mean that calling cmake with `-DCMAKE_C_FLAGS="-fpch-preprocess"` will make CMake discard the value it sets by default for CMAKE_C_FLAGS, same for CXX.
From your link: "For each language, if this variable is not defined, it is initialized and stored in the cache using values from environment variables in combination with CMake's builtin defaults for the toolchain.".
https://cmake.org/cmake/help/latest/envvar/CFLAGS.html#envvar:CFLAGS, the default value would be obtained also from the environment variable CFLAGS, so anything put inside there would be discarded.
That said, it was just an observation and i'm not sure it would eventually make some practical difference when compiling Sphere for x86 or Arm.

MaxKellermann:
Specifying `-DCMAKE_C_FLAGS` will omit any flags set in the `CFLAGS` environment variable, yes, because cmake treats both equal. Is that what you call "its default flags"?

What default flags would be omitted on GitHub actions?

cbnolok:
As i said, probably nothing will be omitted because nothing might be stored in the C(XX)FLAGS env variable in this case, nor anything _relevant_ might be set by default by CMake for a given compiler/build type. It's a personal over-concern about reproducibility and predictability in different contexts, when people might want to just copy/paste what we wrote in the action and compile on their machine

Issue #1249, opened 2024-06-05
Giant Spiders drops i_spider_web in sleeping sector


DavideRei:
Sphere console gives warnings about too many items in different sleeping sectors. When you go there in game the problem disappear.
I found out that the issue is related to spider's webs, giant spiders are able to drop webs in sleeping sector but the timer on the spider web doesnt work because of the sleeping sector.
The problem is solved setting can_o_nosleep on i_spider_web item

Jhobean:
Ok the item are added here: https://github.com/Sphereserver/Source-X/blob/8e444840f7e916126f5f4c1a56b4049e34b20fbd/src/game/chars/CCharNPCAct.cpp#L1918-L1949

A check for sector sleep should be add on spider and fire elemental

Jhobean:
This bug is weird because:
1- In sleeping sector NPC sleep and dont do special action (Just tested and it work)
2- Item with decay flag suppose to always continue timer and decay on sleeping sector (Just tested and it work)

I was about to add some check on the core to prevent
 a: prevent npc to spawn item id sector sleep
b: Add can_o_nosleep flag on item.

But both are useless because   1 and 2  work.....  

I try to reproduce but i'm not able...   I spawn 250 spider on britain and they make spider web everywhere.. I leave for 10 mins,  (Sector sleep after 1 min) When I come back there no web.

Tolokio:
whats the bug then?

Jhobean:
> whats the bug then?

Don't know... what able to reproduce.. Maybe David was using an old build before we added the "decay" don't sleep

Tolokio:
nolok confirmed, so there got to b something.

cbnolok:
Well it was confirmed back then, i didn't check with latest builds

Soulless-1:
it seems to be along the edge of sleeping sectors. from what i recall. i did the same fix as david can_o_nosleep

cbnolok:
Is that still happening with a recent build, such latest dev one? This bug sounds weird to me since i recall we fixed a similar (if not identical) issue.

DavideRei:
I removed can_o_nosleep from the web script, so I'll see if the problem persists

DavideRei:
It seems solved, web spider not increased in 20 hours.

DavideRei:
The bug is back on Linux

![image](https://github.com/user-attachments/assets/82a1c85d-7ded-4cf8-b45e-fbe3206ab096)


[...]

PR #1382, opened 2025-03-05
Fixed: spiderwebs accumulating in sleeping sector (Issue #1249)

cbnolok:
Fixed: If Giant Spiders and Fire Elementals are denied going into sleeping state into a sleeping sector, their dropped items (ie. i_spider_web) went into sleeping state and didn't decay, leading to accumulation (Issue #1249).
  Also, stamina wasn't consumed, thus NPCs had no limit in placing those
items.


Jhobean:
If I read the fix correctly, If the NPC have the flag can o nosleep, you set this falg to the item too?
On my side, spider do not have this flag and they dont sleep... I think bug is more deep than that. I think they sleep but thia mothod of special action is launch anyway.

cbnolok:
This should do now

PR #1398, opened 2025-04-18
Added new ini option to display more bounce messages

mtwango:
Some of the bounce messages were disabled, so the output wouldn't be too verbose, but it could be useful to be able to display them (for example when drinking potion / alcohol, that creates something different other than empty bottle).


Jhobean:
You saying actually there no message when I bounce item? I was certain I see one.😂

Issue #1400, opened 2025-04-19
[Feature Request] CLI switch for ini path


mtwango:
Add switch that would define, where ini files are. I would go with `-S` (settings) or `-I` (ini), since `-C` (as config) is taken.

I guess it would be defined in `CServer::CommandLine`, but it would have to be parsed elsewhere, before loading actual ini files. During `CServerConfig::LoadIni` and ` CServerConfig::LoadCryptIni` it would check for input from switch or fallback to directory of sphereserver binary/excutable `./`

This would help with [windows service bug](https://github.com/Sphereserver/Source-X/issues/1399) (the part with directory change could be removed, the path switch would be added as service parameter during creation) and it should enable placing config files on linux systems to `/etc/` directory, where config files usually reside.

mtwango:
And maybe for pid file also, so it can be placed to `/var/run/` on linux.

mtwango:
Yeah I've got Windows part working, but on Linux I had issue with similar to https://github.com/Sphereserver/Source-X/issues/1403 so I had to investigate.

The bigger reason for this is that I'm trying to make "linux server package". It would compile sphere to debian package, that would create dedicated user, install service files, prepare default account and get ini settings to /etc/ folder, because when something changes there, it will popup action window what to do, similar to one bellow:

```
Configuration file '/etc/sphereserver/sphere.ini'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
```

JulianUO:
This would be pretty nice to have Sphere shipped in a docker image and get scripts, ini, etc from a mounted volume.

Issue #1399, opened 2025-04-19
Windows service not working (win 11)


mtwango:
It used to work a while back, can't say if something changed in Windows or code. Starting as service will throw error, since it will be checking for sphere.ini in `C:\Windows\System32` instead of actual sphere directory. I guess the part with changing directory in `WinMain` doesn't work.

Tested on Windows 11 [Version 10.0.26100.3775]

mtwango:
I've traced the issue here: https://github.com/Sphereserver/Source-X/commit/b9cf3e39194b5d9ad407573e5d89c613d9296bdc

And its not the directory change. I guess it has to be something with the linker forgetting to link stuff.

cbnolok:
Do you have an output error, an exit code or anything to further investigating it?

mtwango:
During service installation `SphereSvrX64_nightly.exe -k install` I get error: 
```
AppName SphereSvrX64_nightly.exe 
  AppVersion 1.1.0.4032 
  AppTimeStamp 681ee876 
  ModuleName ntdll.dll 
  ModuleVersion 10.0.26100.3912 
  ModuleTimeStamp bae89f93 
  ExceptionCode c0000005 
  FaultingOffset 0000000000062673 
  ProcessId 0xe974 
  ProcessCreationTime 0x1dbc3d000654aae 
  AppPath C:\sphereX\SphereSvrX64_nightly.exe 
  ModulePath C:\WINDOWS\SYSTEM32\ntdll.dll 
  IntegratorReportId 7dfcc54f-8123-498f-9add-ee00f8bc5908 
  PackageFullName  
  PackageRelativeAppId  
```

Which I think is related to #1403, but service gets intalled. Log file contains:

```
08:26:Installing Service. (null)
08:26:Install OK Test (Sphere X)
```

When the service is started, it immediately shuts down with errors: `A timeout was reached (30000 milliseconds) while waiting for the SphereServer - Test (Sphere X) service to connect.` (Event ID 7009)

and second error from log (Event ID 7000): 
```
The SphereServer - Test (Sphere X service failed to start due to the following error:
The service did not respond to the start or control request in a timely fashion.
```

The log file gets created in C:\Windows\System32\sphere2025-05-13.log with content:

```
Log date: 2025/05/13 08:30:43
08:30:ERROR:Can't find file 'sphere.ini' in any of the expected paths!.
```

[...]

PR #1407, opened 2025-04-29
Add command line switch for changing ini path (#1400)

mtwango:
Not sure, if everything is done correctly, but it allows usage with cli argument `-I=/path/` to define path to folder with ini files.

See changelog entry for usage and caveeats.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -4018,8 +4018,14 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 	// Display more bounce messages when drinking potions, fishing, crafting etc. (Old sphere behaviour).
     VerboseItemBounce=0
 
-22-04-2024, canerksk
+22-04-2025, canerksk
 - Added: New item function, CARVECORPSE: carves a corpse with SRC = char carving the corpse. 
 - Added: New ini RevealFlag setting REVEALF_ONHORSE, to allow or disallow stealth walking if you are on a mount.
 - Added: new ini CombatFlags setting COMBAT_ATTACK_NOAGGREIVED.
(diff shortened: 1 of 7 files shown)


cbnolok:
Interesting idea and thank you for the contribution, i have a couple of questions:
- Have you tried enclosing the -I argument in double quotes? It might occur that the path has some spaces in it.
- What about another argument to set the base folder Sphere checks for relative paths in spheretables.scp? So that you don't have to change the absolute path if you move the server bundle.
- What about a CServer:: method to check for path overrides arguments in a single place and just call it from ntservice.cpp and spheresvr.cpp, instead of writing (and having to maintain) the same code twice?

mtwango:
- `I` argument with double quotes was working ok on Linux, but wasn't on Windows, it should be fixed now. Tested with paths bellow:

```
Linux:
/home/mulambo/te\ st/SphereSvrX64_debug -I=/opt/sphere\ server/
/home/mulambo/te\ st/SphereSvrX64_debug -I="/opt/sphere server/"

Windows:
C:\GIT\Source-X\Debug\bin-x86_64\Debug\SphereSvrX64_debug.exe -I="C:\sphere test\"
```
- Absolute paths in **spheretables.scp** are actually not required. I've had bad testing data. Tested with defined files, files in root scripts folder and subdirectories. I've removed warning note from changelog.
- I was thinking about it, but since it is a bit different now (however the double quotes stripping shouldn't do any harm on linux) I would keep it separate. Main reason for this is, that currently windows service is not working (#1399) (at least on Windows 11). Even when installed via `-K` param, sphere still looks for `sphere.ini` in `Windows\System32` folder. With the `-I` argument, the service can be installed and create `-I=PATH` to directory, where it was installed from. And then I think the `ntservice.cpp` could be cleaned up so it would call `spheresvr.cpp` first, thus reading ini from there and it could be removed from `ntservice.cpp`. But it is a separate issue. But I could be wrong, this was just my train of thoughts.

mtwango:
Str_CopyLimitNull it is (unqote is not needed, since its just sphere.ini / sphereCrypt.ini file without any directories).

PR #1412, opened 2025-05-04
OF_PetBehaviorOwnerNeutral - missing code

GladieUO:
During merge, the code got probably lost.
PR #1319 
https://github.com/Sphereserver/Source-X/commit/fc6c6cde151834c82f39565a40653abd05ed24e2

### DIFF
--- a/src/game/chars/CCharNotoriety.cpp
+++ b/src/game/chars/CCharNotoriety.cpp
@@ -186,7 +186,7 @@ NOTO_TYPE CChar::Noto_CalcFlag(const CChar * pCharViewer, bool fAllowIncog, bool
 				}
 			}
 
-			if (NPC_IsOwnedBy(pCharViewer, false))	// All pets are neutral to their owners.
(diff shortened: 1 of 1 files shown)


cbnolok:
Yes, but shouldn't it be
`if (IsSetOF(OF_PetBehaviorOwnerNeutral) && NPC_IsOwnedBy(pCharViewer, false)) // All pets are neutral to their owners.`?
Without the !

GladieUO:
> Yes, but shouldn't it be `if (IsSetOF(OF_PetBehaviorOwnerNeutral) && NPC_IsOwnedBy(pCharViewer, false)) // All pets are neutral to their owners.`? Without the !

I dont think it should. We would have to then specify what notoriety to change to, and then else to the neutral notoriety. This way we have false only in case of flag disabled or viewer not owner.

If flag is not enabled, and viewer is owner -> show notoriety neutral
If flag is enabled, and viewer is owner -> show real notoriety
If viewer is not owner -> show real notoriety

cbnolok:
You're right!

Issue #793, opened 2021-10-26
Items are out of containers boundaries when dropping them into a container 


a1exp:
This is not reroducible with Orion (I guess it has additional client-side checks), but reprodicible with the default client. 

First issue. 
When you drop an item to a bag (just drop on a bag witout opening a bag), or create an item inside container (e.g. add gold to player with gold += 5000000 and check gold piles in the bank). Items may be out of container boundaries. See screenshot. 
![image](https://user-images.githubusercontent.com/31130012/138882972-c34b85ad-0211-4c45-ac04-ed0fb0eb2a1f.png)

Second issue.
If you update TDATA3 and TDATA4 of containers then creating items via scripts (e.g. same test with gold += 5000000) will always place items at minX,minY. Seems like `GetRandContainerLoc()` is not called in such cases. 

I fuess this function is bugged https://github.com/Sphereserver/Source-X/blob/12224eef08a5920c7e5248dec4b77fe76b777261/src/game/items/CItemContainer.cpp#L507 since it only calls `GetRandContainerLoc()` when initial coordinates are invalid (e.g. = 0 which is the case for TDATA3 and TDATA4 not initialized). 

Fixing the first issue is more important, imo.

mtwango:
The second part is fixed on dev branch.

The first part can be fixed by defining tdata3/4 on containers. The boundaries by default are greater than the actual container gump and the positions are handled by client. If you define the boudaries with tdata3/4, it will not allow client to drop it outside of the area and it will bounce it where it can be,

PR #1414, opened 2025-05-08
Fix dropping items to container with tdata3/4

mtwango:
This should fix part 2 of #793. Items got always 1,1 position when dropping onto container and condition checking the localtion wasn't run (sice dropping would give -1,-1 and was rewritten).

Part one (item clip) should be fixed by definining container size with tdata3/4 in scripts.

Maybe move this part from code to scripts and deprecate gettings these values with some fallback in future to 0,0,512,512 (as it is in `CItemContainer::ContentAdd`)?

 https://github.com/Sphereserver/Source-X/blob/0911dd22b145d4bb006c9a3371b658ec63df26ca/src/game/items/CItemContainer.cpp#L373

### DIFF
--- a/src/game/items/CItemContainer.cpp
+++ b/src/game/items/CItemContainer.cpp
@@ -466,6 +466,8 @@ CPointMap CItemContainer::GetRandContainerLoc() const
 	const CItemBase *pItemDef = Item_GetDef();
 	GUMP_TYPE gump = pItemDef->m_ttContainer.m_idGump;	// Get the TDATA2
 
+    const int iRandOnce = g_Rand.GetValFast(UINT16_MAX);
+
 	// check for custom values in TDATA3/TDATA4
 	if ( pItemDef->m_ttContainer.m_dwMinXY || pItemDef->m_ttContainer.m_dwMaxXY )
 	{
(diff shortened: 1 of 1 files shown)


cbnolok:
What about having a fallback to the default hardcoded values, in case tdata isn't defined? With a warning message in console, of course

mtwango:
Sure, could be that way. 

I was thinking about new method `CEventLog::EventDeprecated` with some yellowish color. It would print to the console, that this behaviour is deprecated  and an alternative can be found in some place (another method, script, variable, etc.). We could pass some date or version, when it will be removed from source (and replaced with some fallback with error instead of deprecation).

But I would say that it shouldn't be part of this PR as it is beyond fixing #793.

cbnolok:
Yes we can discuss about that new method in a separate issue. I wonder though how often would we use that

mtwango:
I've created separate issue for the deprecation discussion: #1419 so this could be merged.

mtwango:
I would leave it to diffferent one. There is warning if tdata4 is undefined / lower than tdata3.

For the undefined for both would be nice to have it in script pack, which I haven't done.

PR #1417, opened 2025-05-09
Feature: script for creating Debian package

mtwango:
Add option to build deb package from source. Preconfigured for Linux x86_64 Nightly. Cmake pamars has been taken from `.github/workflows/build_linux_x86_64.yml` (except ccache).

Description how to build, install, update and remove included in readme.


cbnolok:
That's great! What about moving this in another super-folder with a name like "deploy"? That could be useful also for future subfolders, like docker config files, rpm scripts, etc

mtwango:
Yeah, I think it could be done. 

I put it in root directory, since if you want to distribute packages in official debian repositories, you need to use their Gitlab and CI requires proper structure. You can check it, for example: https://salsa.debian.org/apache-team/apache2 

The packaging can be kept elsewhere, but it is always easier to mirror repository, than running packaging and keeping upstream branch synced. However, if this is not planned, I can move debian directory elsewhere, .gitnore it in root folder and update packaging script, so it gets copied before running packaging.

Issue #1391, opened 2025-04-04
Inconsistent Layer in @ItemEquipTest and @EquipTest triggers


mtwango:
When trying to equip item and check layer, different methods return different values.

Tested on i_platemail_leggings with Sphere X rev 3941 and clients 2.0.0 + 7.0.107

When doubleclicking to equip:
```
ItemEquipTest Serial: 040001fe5, Layer: 4
EquipTest Serial: 040001fe5, Layer: 4

ItemEquip Serial: 040001fe5, Layer: 4
Equip Serial: 040001fe5, Layer: 4
```

When dragging item to Paperdoll or using **dress** macro:
```
ItemEquipTest Serial: 040001fe5, Layer: 31
EquipTest Serial: 040001fe5, Layer: 31

ItemEquip Serial: 040001fe5, Layer: 4
Equip Serial: 040001fe5, Layer: 4
```

cbnolok:
I'd say this is normal, since 31 = LAYER_DRAGGING. You should have also another trigger firing with layer 4, after you dragged the item on the paperdoll, isn't this the case?

mtwango:
After some digging: I'm pretty sure its packet `0x13` (PacketItemEquipReq)
https://github.com/Sphereserver/Source-X/blob/0911dd22b145d4bb006c9a3371b658ec63df26ca/src/network/receive.cpp#L518

The `itemLayer` should not be used (at least not from packet). This data is sent from client. I'm not sure, where client sends this data from (probably tiledata), but it can be forged (ie equipping legs on head layer) by either sending packet by itself or by modifying client data). It should be loaded from the actual item on server.

The dragging layer should be unset, and actual layer should be set after `client->ClearTargMode(); // done dragging.`

PR #1421, opened 2025-05-15
Fix posibly forged layer and different layer during equip trigger (#1391)

mtwango:
- add check for real and sent layer in PacketItemEquipReq (with debug info)
- swap real layer during ItemEquip withing EquipTest / ItemEquipTest trigger

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -3967,16 +3967,16 @@ Added: 'H' shortcut for variables to get the value as hexadecimal.
 6-12-2024, canerksk
 - Added: LOCAL.WOPTalkMode in @SpellCast trigger and ini setting WOPTalkMode.
     In some clients, different operations can be performed based on this. For example, if this talk mode was sent from Sphere as TALKMODE_SPELL, the operation is performed according to the client version. The possibility of changing the talk mode according to the client version used is now provided. By default, TALKMODE_SPELL.
-    
+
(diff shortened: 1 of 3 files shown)


mtwango:
Added changelog entry.

I'll try to elaborate here:

When player dclicks object (Packet 0x06) the workflow for equipping items works roughly this way:
```
PacketDoubleClick::onReceive ->
CClient::Event_DoubleClick ->
CClient::Cmd_Use_Item ->
CChar::ItemEquip
```
And the object id is found by `uid.ObjFind()` (uid is sent by client).


During equipping via dragging to paperdoll or dress macro (Packet 0x13), its like this:
```
PacketItemEquipReq::onReceive ->
CChar::ItemEquip
```

In `PacketItemEquipReq::onReceive` we are finding object id by checking it in LAYER_DRAGGING `source->LayerFind(LAYER_DRAGGING)` and comparing it to id sent by client. The dragged object has modified layer, since it is currently 'equipped' there. But the EquipTest trigger had this modified layer instead of real / defined one (where it is going to be).

The actual layer during equipping is changed in `LayerAdd(pItem, layer);`, but it was **after** the EquipTest trigger.,

I'm pretty sure, this is very specific behaviour and it could have been worked around by numerous ways (ie using findlayer / findid), but it should unify the equiping behaviour.

PR #1426, opened 2025-05-20
Clamp item onto container edge, when placed out of bounds

mtwango:
…aviour)


cbnolok:
Could you use lesser than 0 instead of - 1? Might be more robust

cbnolok:
Yes i thought so, but it's easy to do a fast change in other parts of the code and cause subtle bugs.
Std clamp is maybe more elegant but absolutely your choice!

mtwango:
Updated both.

PR #1464, opened 2025-09-18
CChar: added ResetVisualRange method

nightNR:
To reset visual range in case supporting client versions before and after VisualRange support

When firstly log in using client supporting VisualRange option and the max view distance is set to > 18 CChar will set `m_iVisualRange` to this value. Then after logging in using client with max distance 18 (f.e. 2.0.0) the VisualRange packet is not sent thus the value is not set and therefore this client is not displaying incoming objects. Resetting this value after selecting character fixes this bug.


cbnolok:
Good idea, though i see some potential issues i'd like to discuss with the currently proposed implementation. This is resetting the visualrange to the default value at login everytime for everyone with any client version, right? This will force change the value, which could be undesired.
Maybe another way would be to reset visualrange (which i'd do with SetVisualRange and not directly) only if the client version requires it, or even make getvisualrange virtual and add a CClient::GetVisualRange which checks every time the client version

nightNR:
> Good idea, though i see some potential issues i'd like to discuss with the currently proposed implementation. This is resetting the visualrange to the default value at login everytime for everyone with any client version, right? This will force change the value, which could be undesired. Maybe another way would be to reset visualrange (which i'd do with SetVisualRange and not directly) only if the client version requires it, or even make getvisualrange virtual and add a CClient::GetVisualRange which checks every time the client version

It is not possible to check the client version every time, because it does not depend only on the client version → some clients allow changing the visual range through configuration. The flow is as follows:

- The client sends a SetVisualRange packet with the configured value, e.g. 22.

- The server compares the provided value with its own configured value, e.g. 24, and assigns the lower value to the CChar instance (22).

- The server returns a SetVisualRange packet to the client with the stored value (22).

- The client applies the provided value as its visual range (22).

Older clients without visual range support do not send a SetVisualRange packet, so the visual range value on CChar never changes.

The problem with your proposed solution is that when you call the SetVisualRange method on CChar, it automatically triggers sending a SetVisualRange packet to the client.

With older clients, this works correctly. But clients with visual range support process this packet and set their visual range to that value, e.g. 18.
After that, the standard flow takes place—but the client now sends back the changed value of 18, not the value from its configuration.

PR #1474, opened 2025-09-24
Restore out of bounds checks in SetAdjacentSectors

DavideRei:


cbnolok:
Out of curiosity, did you find when those checks were removed (maybe by me) with a commit? To be sure those are not redundant and the same conditions checked in a earlier function call

Issue #469, opened 2020-05-10
Verifying account with wrong password/ban/no account


criminalx:
When I try to login with a wrong password, no account, banned account, wrong encryption, it always says: verifying account

https://puu.sh/FILLC/72698dffaa.png

Tested with client 203 and 509

GitMadKing:
i can confirm.  see from lot time ago.

alexdan2:
I make some test:
- ClassicUO client not get any problem
- Client 7.0.10.0 also not any problem

Possible happen only on very old client

alexrmrm:
Its only working for clients up to 7.

i maked test comment that line:

> //if ( GetNetState()->m_clientVersion || GetNetState()->m_reportedVersion )	// only reply the packet to valid clients

and now giving error on 203 of bad password.

xwerswoodx:
After a bit checking, the reason of the issue is really simple, it's because 0xEF packet only sent by 6x+ fully, so older clients doesn't send client data in pre login. So m_reportedVersion never setted for older versions.

I agree the alex solution, because I don't think we need to avoid sending reply that packet comes from client's itself. If some clients (which is made by coders not uo's itself) has issue with this packet, they should care of it. Honestly I don't understand why the developer restricted the reply with 6x+ clients or KR version clients.

As we see information in here; https://docs.polserver.com/packets/index.php?Packet=0xEF
Client datas only sent by KR and Latest 6x+ clients, so using that packet to restricting reply causes this issue, cancelling this line may be the best solution.

mtwango:
Implemented on dev

PR #1479, opened 2025-09-29
Send login rejection to all client requests

mtwango:
Related issue: #469 

Older clients doesn't register its version, but can read rejection message, so we should send it to everybody. Tested on UO client 2.0.0, 7.0.110.48, Classic UO 1.1.0.201, Enhanced Client 4.0.110.39 (see img bellow).

obrazek


mtwango:
Changelog updated

Issue #1422, opened 2025-05-18
Custom Item with can=can_i_damageable error


GladieUO:
Once you place your custom item that has can=can_i_damageble on the ground, the error will show up.
`15:08:ERROR:(Blighted_Grove.scp,495)Point(-1,-1): trying to get a sector for point on map #0 out of bounds for this map(7168,4096). Defaulting to sector 0 of the map.`

```
[ITEMDEF i_blighted_pool]
ID=45852
NAME=Blighted Pool 
TYPE=t_normal
CAN=can_i_damageable

ON=@CREATE 
HITS=100
MAXHITS=100
MORE2=1
ATTR=attr_static
COLOR={254 1 763 1 1441 1 1801 1 2126 1 52 1 2002 1 661 1}
```

mtwango:
Fixed on dev

PR #1482, opened 2025-09-30
Fix can_i_damageable creation

mtwango:
- check, whether CCItemDamageable has been placed in the world before calling tick (#1422)


cbnolok:
It might be worth adding that assert inside GetLink, yes, what do you think about it?

Issue #1478, opened 2025-09-28
Custom housing broken again


GladieUO:
1) during building, something happens and you cannot remove anything anymore.
2) you build second floor, commit the build, and whatever you built, will change to design of first floor.
3) deleting tiles stops work and you need to re-do everything
4) buttons like revert, backup, commit sometimes doesnt do anything
5) building stairs, will add single stair tile above the built stairs

<img width="892" height="759" alt="Image" src="https://github.com/user-attachments/assets/d6e55fc1-f7b0-412c-90dc-419ca7c91487" />

https://ctrlv.tv/Dhiz
https://ctrlv.tv/5JXD
https://ctrlv.tv/FiNI

cbnolok:
I don't remember recent changes to the housing system, could you track the first commit having this issue?

GladieUO:
> I don't remember recent changes to the housing system, could you track the first commit having this issue?

It might be durinbg optimalisation or something, there has been few changes in last months. I can try to indentify.

https://gyazo.com/fdf79bf6bfc685e1a84c73cd3396d4b7

GladieUO:
I tried some debugging with VS, and so far I would say this part is weird.... Additem, even tho I added only one item, is being called many times, and is adding design from ground floor to the second floor.

It is not visible, until you click commit, then part of the first floor shows up in the second floor.
you cannot delete the copied part directly, however by removing the original parts you built in second floor, this will also remove copied parts linked to it. 😁

GladieUO:
Last working version/commit in History of CItemMultiCustom.cpp is this one. https://github.com/GladieUO/Source-X-DEV-/commit/86f05a22d5051049f9ace987991137e65ae4d20c

PR #1484, opened 2025-09-30
Fix of custom multi copying first level into other higher ones

GladieUO:
So i compared the old code with the new one, and there seems to be some mismatches with n16_narrow_n8 and n32_narrow_n8_checked making it dont work.
I also added one condition that was missing, and one extra line.

No idea if its proper fix, however it works.

https://github.com/Sphereserver/Source-X/issues/1478

### DIFF
--- a/src/game/items/CItemMultiCustom.cpp
+++ b/src/game/items/CItemMultiCustom.cpp
@@ -646,7 +646,7 @@ void CItemMultiCustom::AddStairs(CClient * pClientSrc, ITEMID_TYPE id, int16 x, 
         if (!pMultiItem->m_visible)
             continue;
 
-        AddItem(nullptr, pMultiItem->GetDispID(), x + pMultiItem->m_dx, y + pMultiItem->m_dy, z + n16_narrow_n8(pMultiItem->m_dz), iStairID);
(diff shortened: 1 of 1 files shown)


cbnolok:
Thanks for the research and good finding!
Could you check if the problem lies only in the last cast (in GetPlaneZ)? The other casts look legit and i prefer having a widespread use of the named casts to get an immediate idea both of the input and output types.
Also, `naword wPlaneBuffer[PLANEDATA_BUFFER]{}` declaration should already zero-initialize the array (because of the ending `{}`), so memset should be redundant.

GladieUO:
I tried like 10 different variants, but nothing just this worked. Removing the memset breaks this too, just tried that. Also tried just ot be sure only the GetPlaneZ but nothing.

cbnolok:
Thank you, i'd ask you the last favor: could you log the values of both the named cast and the static cast? To see where and how they do differ

cbnolok:
Okay it looks like zero-init via `naword wPlaneBuffer[PLANEDATA_BUFFER]{}` doesn't work because of silly C++ rules (might not be considered as a POD/aggregate type because of inheritance and custom constructors/disabled templated assignment operators). Memset is the way to go then, merging

Issue #1442, opened 2025-06-28
INTERRUPT=0.0 not working again


rodBiase:
Hi, I don't know how long this feature stopped working as I've been away from UO for a bit, but my goal is to adjust the pvp to no disturb the spells. I remember achieving it in 2022, simply changing the INTERRUPT=100.0,100.0 from spells to INTERRUPT=0.0,0.0

Tested with the latest SphereX and script pack from the website, no modifications.

To reproduce: change your greater heal spell to INTERRUPT=0.0,0.0
Cast greater heal with one character, cast magic arrow on the character that is casting greater heal with another character. You will see that the INTERRUPT works just fine as will not interrupt right away. But by the end of the cast (after the spell delay), it will trigger Abort and give you the message the spell fizzles.

I also tried @SpellInterrupt trigger because if I could work with it it would be easy just to set a return 1 there but it is not working as well. 

The interrupt works on sphere D from Coruja.

GladieUO:
Works fine for me.
Try INTERRUPT=0

rodBiase:
> Works fine for me. Try INTERRUPT=0

Already have. Can you tell me which build are you using?

GladieUO:
Compiled at Jun 11 2025 (12:58:18) [branch N/A / build 4096 / GIT hash 632c4e6ac8aecd8f28e43f867d8315cbffa2bf62]

mtwango:
The issue is not spell interrupt, the spell is not interrupted. You can test it by setting spell cast to 10 seconds, start casting and dmg yourself with  another character. It will fail ontick after timer is done.

The spell fails, because skill fails during success stage, because the spell id bellow is somewhere set to zero https://github.com/Sphereserver/Source-X/blob/ecd546ab0557d78d5512c791261370734473294b/src/game/chars/CCharSkill.cpp#L2989

mtwango:
Should be fixed on dev (#1559)

PR #1486, opened 2025-10-01
Add EF to walk over monsters

mtwango:
optional OF_DontInterruptHealOnHit to skip interrupt on getting hit, so scripters can implement their interrupt in hit / gethit triggers
experimental EF_WalkBypassMonsters to leverage personalspace trigger on NPC without full stamina. Experimental, because this is resource intensive (lots of packets) when denied by trigger.


mtwango:
Note: healing doesnt get interrupted by hit, but still fails on skill abort. Might be related to #1442

mtwango:
The healing OF is for discussion.

Maybe it should not be as a OF, but removed by default. It doesn't take into account veterinary (in core) and seems like an agressive way to stop action, since scripters can make an easy replacement in scripts, ie in character events:

```c
on=@getHit
  if (<action> == Healing)
    sysmessage <defmsg.healing_interrupt>
    action = -1
  endif
```

mtwango:
Great discussion :D

I've updated the PR to remove healing interrupt and how to add it to script pack.

PR #1493, opened 2025-10-06
Fix of pets not continue fight after casting spell (All kill commands)

GladieUO:
So there has been long time issue of pets that suppose to be casters, not continue fighting after spell gets off. They would immediately follow master.
Turns out whenever the master doesnt have memory OnHarmedBy(Enemy), pets will just not stick to the target.

If player wanted his pets/summons to stick fighting and casting, he would have to call all guard me and attack target.


Jhobean:
I like this fix. Maybe there a better way to achieve it but it seem effective. I dont see what negative impact it could have.

Issue #1458, opened 2025-09-11
Fizzle Effect & Fizzle Sound


xMirroRx:
Magery Fizzle Sound Effect
Magery Fizzle Effect ID
It would be great to be able to change these with an Sphere.ini support.

mtwango:
local.sound added, see #1520 comment how to overwrite global / local behaviour

xMirroRx:
It's very good on my end, thanks.

PR #1520, opened 2025-11-23
Sound and item effects for failed spells have been added to the spher…

canerksk:
…e.ini setting #1458


mtwango:
I've added local.sound to spellfail/fail triggers (and credited you in changelog). 

You can overwrite global behaviour in spellfail trigger (for players and npcs) and local in each spell fail trigger, no ini settings needed.

Issue #1189, opened 2023-11-22
Human Corpse


AtillaG1:
If you kill human, either player of NPC with hair and beard, the corspe shows 2 items as lootables in default, even tho you cant see them.

xwerswoodx:
it's not a bug, technically you see them, Hair and Beard have to move into  the corpse to make corpse look like you. I meant beard and hair is already on the corpse, you can easily see them.

![image](https://github.com/Sphereserver/Source-X/assets/6155640/a3af1c4c-ac32-4c66-9076-7a7ba1aa62f7)

Jhobean:
I think he mean in the name we see 5 items and there only 3 items lootable.

xwerswoodx:
Uhm, we could make newbie items ignored when you check the resource count in corpse, but need to make some check for it.

PR #1523, opened 2025-11-25
Fixed the issue where some items (hair, beard, etc.)

canerksk:
#1189


mtwango:
The items with attr `ATTR_CURSED2`, `ATTR_BLESSED2` and `ATTR_STATIC` can be seen in the body (as player and gm). Those shouldn't be skipped.
{96EE5BA0-EF21-47B4-A975-EFBFC0AE5B0D}

PR #1531, opened 2025-12-05
Rework github worflow Replace TAG and add a Dev release

Jhobean:
I reworked the Workflows. This bring 2 features
1- delete the Tag each time  and update all the date and sources files
2- Add a release for dev branch.

Tested and work Ok on my fork

### DIFF
--- a/.github/workflows/build_linux_x86.yml
+++ b/.github/workflows/build_linux_x86.yml
@@ -120,7 +120,7 @@ jobs:
     # Upload artifact linked to the workflow run - only if the run is for a pull request, or for selected branches
     - name: Upload artifact
       if: contains(fromJson('["master", "main", "dev"]'), github.ref_name) || github.event_name == 'pull_request'
-      uses: actions/upload-artifact@v4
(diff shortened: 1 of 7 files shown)


mtwango:
Few notes:

- Can you check the permissions on the actions? Some I think are very loose (write) and some actions are missing actions.
- The tag should have different name (like development). Otherwise you will have ambiguous name `dev` for tag and branch
- The condition for coverity scan should be better done checking if the secret is set (not empty), so forked repos can run their own covscan, if they have their keys
- Why are you resseting the tag? To show the latest time on tags page? Wouldn't it be better to use `move_existing_tag: true` in gh release action?

Also, please keep the changes clean. As I understand it, it was generated with AI, but its hard to see what changed (new stuff, changed stuff) and what was just refactored (some blocks moved, whitespaces added and removed). Thank you

Jhobean:
The purpose about the tag and write mode was necessery because i was delete the old tag and recreate one. The purpose is to have the correct date write next to the version. 

I recheck after the spring break.

Issue #1539, opened 2026-01-04
Casting & different monster dying aborts the cast on player


GladieUO:
we found an interesting bug, if you cast spell on monster A, meanwhile monster B dies during the cast time, the ACTARG1 on the player changes to 00 and the cast is @abort because the spell id is stored in ACTARG1 (m_atMagery.m_iSpell). The bug is that somehow the ACTARG1 is erased by monster closeby dying.

So far it only happens if the dying monster is attacking you, if not, it doesnt happened.

mtwango:
Should be fixed on dev (#1559)

Issue #1505, opened 2025-10-25
Shrinked pets deleted after GC


GladieUO:
There is some bug with the shrinking of pets and garbage collection.

Its very rare, but annoying. You shrink you pet -> and here I suspect something happenes, because next time the GC is ran before save, it will flag the figurine or the disconnected char as problematic or non existant.
Causing you to unshrink the pet, spawning completely new one, only NAME is kept.

```
22:07:ERROR:GC: Deleted UID=0b7a2, Defname='c_spider_dread_tame', Name='Webby'. Invalid code=0x1204 (Ridden NPC without a valid mount item).
22:07:ERROR:Garbage Collection: done. Object memory leak 166922!=166916.
```

```
02:38:ERROR:GC: Deleted UID=070cb, Defname='c_allosaurus_tame', Name='Dino'. Invalid code=0x1203 (Ridden NPC not acting as such).
02:38:ERROR:GC: Deleted UID=04001c54b, Defname='i_pet_allosaurus', Name='Dino'. Invalid code=0x2104 (Disconnected item).
```

```
02:52:ERROR:GC: Deleted UID=040004c5d, Defname='i_memory', Name='Koppak'. Invalid code=0x1103 (Ridden NPC not acting as such).
02:52:ERROR:GC: Deleted UID=0162f6, Defname='c_allosaurus_tame', Name='Koppakuv A'. Invalid code=0x1203 (Ridden NPC not acting as such).
02:52:ERROR:GC: Deleted UID=040031c42, Defname='i_pet_allosaurus', Name='Koppakuv A'. Invalid code=0x2104 (Disconnected item).

```

mtwango:
Should be fixed on dev (#1559)

PR #1540, opened 2026-01-05
Fix casting abort

GladieUO:
Casting was aborted if other monster with attacker (me) died. Because Attacker_Clear nulls the ACTARG1 (which is spell def during cast). #1539


GladieUO:
It might be also related to the pets bug, where the ACTARG1 on pet is being anulled and thus the link to its figurine is gone.
It might proabbly happend in moment when you shrink pet, but at same time the monster dies and triggers attacker_clear on pet being shrinked.
#1505

mtwango:
Targeting should be fixed (#1559)

Issue #1537, opened 2026-01-01
Changing body stores wrong body and damage


mtwango:
When changing character body, the new body doesn't load chardef properly. For example:

```
[CHARDEF c_test]
ID=027
NAME=Vampire Bat
DAM=14,36
ARMOR=56
.....
```

Using `.setbody c_test` on player character, it changes its armor, but it desnt change damage. When using `.xshow body`, it shows `027` (or `c_mongbat`) instead of `c_test`

mtwango:
The body changing part is here and it should probably be `SetID()` instead of `SetDispID()`: https://github.com/Sphereserver/Source-X/blob/5d62a1ebc566002b936eead17835f5a207211315/src/game/chars/CChar.cpp#L3088

mtwango:
The workaround for dmg would be in `CChar::Fight_CalcDamage` using 
```
const CCharBase *pChar = Char_GetDef();
iDmgMin =  pChar->m_attackBase;
iDmgMax = iDmgMin + pChar->m_attackRange;
```
if character doesn't have a weapon to load dmg from actual body, but it would be better to change dam value for current character, when he changes the body.

It used to work like that, but it was changed in 37350b652d435bd7f2d16d4796d68352d75120c5 (2015)

Jhobean:
I already have a bug about changing body too. 

I used a workaroud https://github.com/Sphereserver/Source-X/issues/1374

Maybe it could give hint for permanent fix

PR #1541, opened 2026-01-13
Use GetID for setting BODY (#1537)

mtwango:


mtwango:
Not sure if this is viable. Body, id and baseid would be the same.

PR #1550, opened 2026-02-04
fix: pre-AOS armor damage reduction formula

nightNR:
Fix a probable bug where the random value is chosen between iArMin and iArMax - iArMin, which makes no sense when iArMin = iArMax / 2


mtwango:
Can you make the higher value +1? so the values are not the same when the iArMax is close to `< 2`

PR #1559, opened 2026-03-04
Fix reseting target if other combatant dies or leaves combat

### LINKED ISSUE #1442: INTERRUPT=0.0 not working again

### LINKED ISSUE #1539: Casting & different monster dying aborts the cast on player

mtwango:
Don't reset actarg1 in fight_clear, if our target was different from the target we are clearing the fight with (it causes the current action to fail).

Duplicates / Similar to #1540 

Note: maybe actarg2 and actarg3 should also not be reseting...


mtwango:
Needs testing, but should theoretically fix #1539 #1442 #1505

PR #1579, opened 2026-05-16
Fix the error message when try login with a linger char.

Jhobean:
I used a static_cast because I thought simply enter a 5 was not good. We already beter define  this value on the PacketWarningMessage method


mtwango:
Seems a bit drastic solution, might be better something like this: https://github.com/mtwango/Source-X/commit/edebaa612f0a9b12d84f5ba0ee2969c8258fb647

Jhobean:
True!!!  It work?  Because you choose sucess, the correct error packet will follow after? (Maybe add a comment  explaining why we send sucess packet?)

